PR #3, opened 2020-10-04
Change default namespace to komodor 

shikloshi:


MickaelAlliel:
Don't update the `kube-install` manifests manually
1. Bump chart minor version in Chart.yaml (required anyway, wont be published if not)
2. run `make generate-kube`
3. commit

@shikloshi

MickaelAlliel (approved):

MickaelAlliel:
Small concern about already existing installations. Upgrading namespaces is tricky and should be done as a new install.

shikloshi:
> Small concern about already existing installations. Upgrading namespaces is tricky and should be done as a new install.

I will test the new installation "upgrade" use case today to see that it goes okay and no helm issues there.

PR #7, opened 2020-11-24
Update docs and version

shikloshi:


charts/k8s-watcher/values.yaml
```diff
@@ -3,6 +3,7 @@ replicaCount: 1
 image:
   repository: komodorio/k8s-watcher
   pullPolicy: IfNotPresent
+  tag: "0.1.10"
```
MickaelAlliel, line 6: @shikloshi remove this, update in `Chart.yaml` both the chartVersion and appVersion

MickaelAlliel (changes requested):

PR #8, opened 2020-11-25
update docs and add to example

shikloshi:


MickaelAlliel (approved):

charts/k8s-watcher/values.yaml
```diff
@@ -33,6 +33,7 @@ watcher:
   namespacesBlacklist:
     - kube-system
   nameBlacklist: []
+  redact: []
```
MickaelAlliel, line 36: Update the Chart.yaml `chartVersion` for this to be applied

PR #27, opened 2021-05-03
Allow Komodor api key to be provided by an existing kubnernetes secret

liorrozen:
These changes allow using an existing secret resource for the Komodor `apiKey` instead of needing to specify the token for each helm command or hardcoding it into `values.yaml`.

For clusters that utilize external secrets (such as https://github.com/external-secrets/kubernetes-external-secrets), charts that allow providing an existing secret provide the functionality needed to remove all sensitive information from the chart files themselves.

These changes are an attempt to implement similar functionality that bitnami has in many of their charts. For example. the [redis chart](https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml#L101)

### DIFF
--- a/charts/k8s-watcher/Chart.yaml
+++ b/charts/k8s-watcher/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v1
 name: k8s-watcher
 description: Watches and send kubernetes resource-related events
 icon: https://raw.githubusercontent.com/komodorio/helm-charts/master/placeholder-logo.png
-version: 0.1.39
+version: 0.1.40
 appVersion: 0.1.33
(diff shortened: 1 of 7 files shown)


MickaelAlliel:
Hey @liorrozen !
Thanks a lot for contributing :)
The feature and pull request itself looks good, I would only ask to provide possibility to use existingSecret for `kialiApiKey` value as well, because the if condition in the secret itselfs means kialiApiKey is not created as well in this case when using existingSecret.

I would be more than happy to copy the changes and fix this small issue myself if you'd prefer :) Let me know

MickaelAlliel (approved):

PR #70, opened 2022-01-31
add tag when releasing version

nirbenator:


nirbenator:
CU-221439v-release-when-tag

MickaelAlliel (approved):

PR #83, opened 2022-04-26
mod: task execution and pods and logs resources are now allowed by default

MickaelAlliel:
CU-2ebbc9u


nirsht (approved):
Please change the name of the commit to public so you trigger release note

PR #85, opened 2022-04-26
CU-296c4bk - feat: custom annotations on agent deployment

MickaelAlliel:


nirbenator (approved):

nirbenator:
@MickaelAlliel  make sure you set a public: merge commit mentioning that this flag is now available so it will show in RN

PR #96, opened 2022-07-24
CU-2me1nfd - Public: add clusterRole permissions for restart service action

Iriskella:


charts/k8s-watcher/kube-install/k8s-watcher/templates/clusterrole.yaml
```diff
@@ -5,215 +5,223 @@ metadata:
...
+      - watch
+      - list
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
+      - replicasets
+      - statefulsets
+    verbs:
+      - patch
```
MickaelAlliel, line 224: What about the helm chart?
You updated auto-generated files here.

You should update the helm templates, and run `make docker-generate-kube` to generate the static kubernetes manifests.
The kustomize manifests have to be updated manually for now as well (in the root of the repo)

MickaelAlliel (changes requested):

MickaelAlliel (approved):

PR #98, opened 2022-07-25
CU-2me6ppn - add permissions for delete command

Iriskella:


Iriskella:
this will be merged with Re-run jobs

PR #106, opened 2022-09-08
CU-2teaez0 - public:  Add pods eviction permissions under action.advanced value

michaeldeift:


nirbenator (approved):
lgtm - add a public note 
public: Adding pod eviction permission when action.advanced is passed to support feature X 
do it on the merge commit :)

PR #128, opened 2023-01-10
Add tmp directory as volume in container

alon-glater:
Done in order to allow for the creation of temporary files on agent, since write permissions were revoked on the agent container for security concerns.


nirbenator (approved):
lgtm

PR #130, opened 2023-01-11
🎨 Remove Superfluous Rule from helm-dashboard ClusterRole

toddtee:
A superfluous rule is added to the ClusterRole upon creation, when the dashboard.allowWriteActions value is set to true. This commit will ensure that only a single rule is created within the ClusterRole, regardless of whether the dashboard.allowWriteActions value is enabled or not. The verbs within this rule will update accordingly.


toddtee:
This commit will address issue #129 .

undera:
The master source for the helm-dashboard chart is in its repository, please open your PR there

toddtee:
Sure thing!

PR #131, opened 2023-01-11
CU-860phernx - feat: pipeline change - wait for chart to be updated

nirbenator:
removed a 30 second sleep waiting for helm repo to update with the new version
added validation that the version was actually updated

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -62,3 +62,27 @@ git add .
 git status
 git commit -m "Published by Buildkite $BUILDKITE_BUILD_URL"
 git push origin "$GITHUB_PAGES_BRANCH"
+
+
+echo ">> verifying chart version update"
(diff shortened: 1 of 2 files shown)


publish.sh
```diff
@@ -62,3 +62,27 @@ git add .
...
 git commit -m "Published by Buildkite $BUILDKITE_BUILD_URL"
 git push origin "$GITHUB_PAGES_BRANCH"
+
+
+echo ">> verifying chart version update"
+helm repo add komodorio https://helm-charts.komodor.io
+helm repo update
+helm repo list
+
+for x in {1..10}; do
```
MickaelAlliel, line 72: I think we might need to do `helm repo update` again in the loop no? @nirbenator

MickaelAlliel (approved):

PR #132, opened 2023-01-11
CU-860pguwm4 Add permission flag for pod exec

undera:


charts/k8s-watcher/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v1
 name: k8s-watcher
 description: Watches and sends kubernetes resource-related events
 icon: https://raw.githubusercontent.com/komodorio/helm-charts/master/k8s-watcher.svg
-version: 1.2.35
+version: 1.3.35
```
MickaelAlliel, line 5: Set 1.3.0 @undera

MickaelAlliel (changes requested):
See version comment

PR #134, opened 2023-01-12
fix pvc storageClassName

Wojteek97:
With the actual pvc.yaml i can't add a storageClassName properly, my change includes the necessary configuration to use a custom storageClassName.


undera:
The master source for the helm-dashboard chart is in its repository, please open your PR there

PR #139, opened 2023-01-15
Validate chart is updated before installing agents

nirbenator:
validation for chart update before we continue to deploy
tested in branch run [here](https://buildkite.com/komodor/helm-charts/builds/799#0185b57e-0667-499f-9bad-40abf8c68e2d)


nirbenator:
CU-860phqbwm

MickaelAlliel (approved):

PR #145, opened 2023-01-30
fix: Make PVC  use storageClass from values

dploeger:


undera:
Master source for this chart is maintained in main repo here: https://github.com/komodorio/helm-dashboard/tree/main/charts/helm-dashboard

undera:
Please create a PR against master repo, otherwise the changes will be overwritten here

PR #148, opened 2023-02-23
CU-860q0eueg: Allow all CR to be fetched

YadgarYossi:
## Motivation
For the new product capability of CR/CRD inspection, we might need read permissions for all the possible dynamic groups.

## Changelog
1. Allow all flag
2. Provide whitelist flag


.buildkite/pipeline.yml
```diff
@@ -74,7 +74,7 @@ steps:
...
+      - helm upgrade --install k8s-watcher-staging komodorio/k8s-watcher --set namespace=komodor-staging --set watcher.serverHost=https://staging.app.komodor.com --set watcher.tasksServerHost=https://staging.app.komodor.com --set watcher.telemetry.enable=false --set apiKey=$$API_KEY --set watcher.telemetry.enable=true --set watcher.collectHistory=true --set watcher.nameDenylist="{leader,election}" --set watcher.resources.secret=true --set watcher.redact="{.*KEY.*,.*key.*,.*BUGSNAG.*}" --set watcher.enableAgentTaskExecution=true --set watcher.enableAgentTaskExecutionV2=true --set watcher.allowReadingPodLogs=true --set watcher.actions.basic=true --set watcher.actions.advanced=true --set watcher.enableHelm=true --set helm.enableActions=true --set watcher.actions.podExec=true --set watcher.resources.allowReadAll=true
```
amit9192, line 77: If we set it to true by default I'm not sure we need this flag in the pipeline

charts/k8s-watcher/templates/clusterrole.yaml
```diff
@@ -423,6 +423,23 @@ rules:
...
+{{- if .Values.watcher.resources.allowReadAll }}
+  - apiGroups: ["*"]
+    resources: ["*"]
+    verbs:     ["get", "watch", "list"]
+{{- end}}
+
+{{- $numOfCustomApiGroups := len .Values.watcher.resources.customAPIGroups }}
+{{- $shouldLimitApiGroups := gt $numOfCustomApiGroups 0 }}
+{{- if and $shouldLimitApiGroups (not .Values.watcher.resources.allowReadAll) }}
+  {{- range $cag := .Values.watcher.resources.customAPIGroups }}
```
amit9192, line 436: Do we really need this complexity? 
What will happen if the customAPIGroups will be defined although the allowReadAll is true?

charts/k8s-watcher/values.yaml
```diff
@@ -113,6 +113,8 @@ watcher:
     csiNode: true
     csiStorageCapacity: true
     volumeAttachment: true
+    allowReadAll: true
+    customAPIGroups: []
```
amit9192, line 117: It is unclear how it should be formatted + need to add documentation on the values table in the md file https://github.com/komodorio/helm-charts/blob/master/charts/k8s-watcher/README.md

charts/k8s-watcher/templates/clusterrole.yaml
```diff
@@ -423,6 +423,23 @@ rules:
...
+  - apiGroups: {{ $cag.apiGroups }}
+  - resources: {{ $cag.resources }}
+  - verbs: {{ $cag.verbs }}
```
erez-rabih, line 439: I think it is not the correct syntax to have apiGroups resources and verb in an array
it should be
```yaml
- apiGroups: {{ $cag.apiGroups }}
  resources: {{ $cag.resources }}
  verbs: {{ $cag.verbs }}
```

also, I think it is enough to have only the api groups in the `customAPIGroups` and use * for resources and get list watch as verbs

charts/k8s-watcher/templates/clusterrole.yaml
erez-rabih: I agree that we can just iterate over the customAPI groups regardless of the read all flag
we also have this duplication in all other native resources

charts/k8s-watcher/README.md
```diff
@@ -153,6 +153,9 @@ The following table lists the configurable parameters of the chart and their def
 | `watcher.resources.csiNode`                        | Enables watching CSINode                                                                                                                                                      | `true`                              |
 | `watcher.resources.csiStorageCapacity `            | Enables watching CSIStorageCapacity                                                                                                                                           | `true`                              |
 | `watcher.resources.volumeAttachment`               | Enables watching VolumeAttachment                                                                                                                                             | `true`                              |
+| `watcher.resources.csiStorageCapacity `            | Enables watching CSIStorageCapacity                                                                                                                                           | `true`                              |
```
erez-rabih, line 156: what is csiStorageCapacity?

.buildkite/pipeline.yml
erez-rabih: yes if the value is true in values.yaml we don't need it here

erez-rabih (changes requested):

charts/k8s-watcher/templates/clusterrole.yaml
YadgarYossi: Thanks for the correction. I am not convinced that we might need to remove the resources/verbs with this current configuration name. Now the name is `customAPIGroups` and not `customReadAPIGroups` so it could confuse that it will be about getting/list/watch, and will set all the resources. WDYT?

charts/k8s-watcher/README.md
YadgarYossi: My mistake, removed

charts/k8s-watcher/templates/clusterrole.yaml
erez-rabih: as I see it we have two options:
1. allow users to list custom api groups (we could name it customReadAPIGroups) and we will add the groups with read permissions on all resources under these groups
2. allow users to freely add rules to the agent RBAC policy - groups, resources and verbs. in that case I would just call the value additionalRBACPermissions and embed the provided YAML as-is into the agent's ClusterRole

YadgarYossi: Fixed.

charts/k8s-watcher/templates/clusterrole.yaml
```diff
@@ -423,6 +423,20 @@ rules:
...
 {{- end}}
+
+{{- if .Values.watcher.resources.allowReadAll }}
+  - apiGroups: ["*"]
+    resources: ["*"]
+    verbs:     ["get", "watch", "list"]
+{{- end}}
+
+{{- range $cag := .Values.watcher.resources.customReadAPIGroups }}
+- apiGroups: {{ $cag.apiGroups }}
```
erez-rabih, line 434: I think we need to indent this clause

erez-rabih (approved):

YadgarYossi:
public: Full support in CRD inspection.

1. Agents will be reported also for custom resources data.
2. Full read (list, get, watch) permissions for all the groups by default - set by `allowReadAll` to true as default.
3. Customization for allowed API groups for read - set by `customReadAPIGroups`

amit9192:
> public: Full support in CRD inspection.
> 
> 1. Agents will be reported also for custom resources data.
> 2. Full read (list, get, watch) permissions for all the groups by default - set by `allowReadAll` to true as default.
> 3. Customization for allowed API groups for read - set by `customReadAPIGroups`

Adding custom resources support.

Permissions - Adding the following values:
`allowReadAll` - provides read permissions on all resources (defaults to true)
`customReadAPIGroups` - allows specifying a list of custom resources API groups (relevant if allowReadAll is set to false)

charts/k8s-watcher/README.md
```diff
@@ -152,7 +152,9 @@ The following table lists the configurable parameters of the chart and their def
...
-| `watcher.resources.volumeAttachment`               | Enables watching VolumeAttachment                                                                                                                                             | `true`                              |
+| `watcher.resources.volumeAttachment`               | Enables watching VolumeAttachment                                                                                                                       | `true`                              |
+| `watcher.resources.allowReadAll`               | Enables reading all the resources in the cluster                                                                                      | `true`                              |
+| `watcher.resources.customReadAPIGroups`               | A whitelist of allowed resources to read from cluster - each array element should a string which represents the group name                                                                                    | `[]`                              |
```
erez-rabih, line 157: 1. this sentence is grammarly incorrect - `each array element should a string`
2. I would rephrase this to `A list of custom API groups to allow read access to`
3. also I'd keep indentation of the `|` mark on this line and on line 156

erez-rabih (approved):

PR #149, opened 2023-03-01
CU-860q0eueg: fix customReadAPIGroups condition to support empty values

YadgarYossi:


MickaelAlliel (approved):

YadgarYossi:
public: Adding custom resources support.

Permissions - Adding the following values:
allowReadAll - provides read permissions on all resources (defaults to true)
customReadAPIGroups - allows specifying a list of custom resources API groups (relevant if allowReadAll is set to false)

PR #151, opened 2023-03-02
CU-860q0eueg: Fix k8s-watcher clusterrole new condition to support older helm versions

YadgarYossi:
## Motivation
The previous `gt` condition syntax was breaking old helm versions and resulted:
`Error: template: k8s-watcher/templates/clusterrole.yaml:433:57: executing "k8s-watcher/templates/clusterrole.yaml" at <gt>: wrong number of args for gt: want 2 got 0`

This is the fix for that.

### DIFF
--- a/charts/k8s-watcher/templates/clusterrole.yaml
+++ b/charts/k8s-watcher/templates/clusterrole.yaml
@@ -430,7 +430,7 @@ rules:
     verbs:     ["get", "watch", "list"]
 {{- end}}
 
-{{- if and .Values.watcher.resources.customReadAPIGroups gt (len .Values.watcher.resources.customReadAPIGroups) 0 }}
+{{- if and .Values.watcher.resources.customReadAPIGroups (gt (len .Values.watcher.resources.customReadAPIGroups) 0) }}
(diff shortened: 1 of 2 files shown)


YadgarYossi:
public: Support the older helm version for the new `customReadAPIGroups` capability

charts/k8s-watcher/templates/clusterrole.yaml
```diff
@@ -430,7 +430,7 @@ rules:
     verbs:     ["get", "watch", "list"]
 {{- end}}
 
-{{- if and .Values.watcher.resources.customReadAPIGroups gt (len .Values.watcher.resources.customReadAPIGroups) 0 }}
+{{- if and .Values.watcher.resources.customReadAPIGroups gt ((len .Values.watcher.resources.customReadAPIGroups) 0) }}
```
erez-rabih, line 433: the brackets are in the wrong place
`(gt len .Values.watcher.resources.customReadAPIGroups 0)`

erez-rabih (changes requested):

erez-rabih (approved):

PR #154, opened 2023-03-14
CU-860q4n8y4 -add charts

Noy-Simon:
public: Add new metrics agent ( not active by default )

### DIFF
--- a/charts/k8s-watcher/templates/configmap-metrics.yaml
+++ b/charts/k8s-watcher/templates/configmap-metrics.yaml
@@ -9,33 +9,34 @@ metadata:
 data:
   telegraf.conf: |
     [global_tags]
-      AccountId = "${AccountId}"
+      account_id = "${ACCOUNT_ID}"
+      cluster_name = "${CLUSTER_NAME}"
 
     [agent]
-      interval = "15s"
-      flush_interval = "1m"
+      interval = "${INTERVAL}"
+      flush_interval = "${FLUSH_INTERVAL}"
 
     [[outputs.timestream]]
-      region = "us-east-1"
-      access_key = "${TimestreamAccessKey}"
-      secret_key = "${TimestreamSecretKey}"
+      region = "${TIMESTREAM_REGION}"
+      access_key = "${TIMESTREAM_ACCESS_KEY}"
+      secret_key = "${TIMESTREAM_SECRET_KEY}"
 
-      database_name = "prod-metrics-db"
+      database_name = "${TIMESTREAM_DATABASE_NAME}"
 
       describe_database_on_start = true
-      mapping_mode = "multi-table"
+      mapping_mode = "${TIMESTREAM_MAPPING_MODE}"
 
       create_table_if_not_exists = true
 
-      create_table_magnetic_store_retention_period_in_days = 37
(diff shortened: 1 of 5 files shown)


charts/k8s-watcher/templates/configmap-metrics.yaml
```diff
@@ -0,0 +1,57 @@
...
+    [agent]
+      interval = "15s"
+      flush_interval = "1m"
+
+    [[outputs.timestream]]
+      region = "us-east-1"
+      access_key = "${TimestreamAccessKey}"
+      secret_key = "${TimestreamSecretKey}"
+
+      database_name = "prod-metrics-db"
```
erez-rabih, line 23: there are many configuration details hardcoded here which might lock us in the future if we'd like to change them
I'm referring to database name, region, retention periods, scrape interval and flush interval etc

IMO we should be receiving these from the collector endpoint for metrics configuration the same way we do with the api keys

not a lot of work but will allow us to be more agile if we need to

charts/k8s-watcher/templates/configmap-metrics.yaml
```diff
@@ -0,0 +1,57 @@
...
+      fieldpass = ["memory_usage_bytes"," "]
+      namedrop = ["kubernetes_system_container"]
+
+    [[inputs.kube_inventory]]
+      ## URL for the Kubernetes API
+      url = "https://kubernetes.default.svc.cluster.local"
+      bearer_token = "/run/secrets/kubernetes.io/serviceaccount/token"
+      insecure_skip_verify = true
+      resource_include = [ "pods", "nodes" ]
+      fieldpass = ["resource_requests_cpu_units","resource_limits_cpu_units","resource_requests_memory_bytes","resource_limits_memory_bytes"]
```
erez-rabih, line 54: does it mean we also pass the limits and requests?
we already have this information in our DB so we might be able to drop it

charts/k8s-watcher/templates/daemonset-metrics.yaml
```diff
@@ -0,0 +1,62 @@
+{{- if .Values.metrics.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
```
erez-rabih, line 3: do we really have to install it as a daemonset on clients? won't a deployment suffice for this matter?

charts/k8s-watcher/templates/configmap-metrics.yaml
Noy-Simon: will do

charts/k8s-watcher/templates/configmap-metrics.yaml
Noy-Simon: we talked about it and decided to collect it for now
we can change it in the future if we want

charts/k8s-watcher/templates/daemonset-metrics.yaml
Noy-Simon: we wanted to support large costumers 
so we decided to use Daemonset

charts/k8s-watcher/templates/configmap-metrics.yaml
Noy-Simon: done

erez-rabih (approved):

PR #159, opened 2023-03-21
CU-860q9q81c - add image pull policy

Noy-Simon:
public: metrics agent image config update


galantaa (approved):
don't you need to add the value?

PR #163, opened 2023-03-23
CU-860qak517 - ImagePullSecret

Moran-k:
Add imagePullSecret to `k8s-watcher` service account.

To use the image pull secret, pass `--set imagePullSecret={secret name}` to the helm command.

If the secret does not exist, the SA will ignore it until it is available.


Moran-k:
public: Added imagePullSecret

nirbenator (approved):

PR #165, opened 2023-03-26
CU-860qa57t5 mounts a read write volume for k8s api cache

galantaa:


galantaa:
public: Added a new rw volume for k8s api cache

charts/k8s-watcher/templates/deployment.yaml
```diff
@@ -122,3 +126,8 @@ spec:
           emptyDir:
             sizeLimit: {{ .Values.helm.volumeSizeLimit | default "256Mi" }}
 {{- end }}
+{{- if .Values.enableRWCache }}
+        - name: kube-cache
+          emptyDir:
+            sizeLimit: 256Mi
```
galantaa, line 132: locally mine is 136Mi, so I guess we can start with that

galantaa:
@MickaelAlliel I installed it locally and I can write to `/.kube/` 🎊

MickaelAlliel (approved):

PR #166, opened 2023-03-28
CU-860qa4tx6 - rename metrics agent

Noy-Simon:
public: rename metrics agent

### DIFF
--- a/charts/k8s-watcher/values.yaml
+++ b/charts/k8s-watcher/values.yaml
@@ -50,6 +50,7 @@ helm:
 
 metrics:
   enabled: false
+daemon:
   resources:
     limits:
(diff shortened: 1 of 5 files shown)


Noy-Simon:
should be deployed close with :
https://github.com/komodorio/komodor-agent/pull/244

charts/k8s-watcher/values.yaml
```diff
@@ -50,6 +50,7 @@ helm:
 
 metrics:
```
galantaa, line 51: replace?

galantaa: where do u use it?

galantaa (approved):

PR #176, opened 2023-04-17
CU-860qjcbjw - copy metrics.enabled config to watcher.daemon.enabled for UI agent installed check 

Noy-Simon:
public: update daemon config value

### DIFF
--- a/charts/k8s-watcher/values.yaml
+++ b/charts/k8s-watcher/values.yaml
@@ -49,7 +49,7 @@ helm:
   volumeSizeLimit: "256Mi"
 
 metrics:
-  enabled: false
+  enabled: &metricsEnabled false
 daemon:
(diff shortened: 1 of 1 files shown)


charts/k8s-watcher/values.yaml
```diff
@@ -71,6 +71,7 @@ watcher:
   allowReadingPodLogs: true
   enableHelm: true
   daemon:
```
erez-rabih, line 73: instead of duplicating the boolean value in two separate places in the config, can't we just refer to the same YAML path when we read the configuration?

erez-rabih: I see thanks for explaining

erez-rabih (approved):

PR #177, opened 2023-04-18
CU-860qjr91h - add config to configmap

Noy-Simon:
public: update configmap


Noy-Simon:
This pr added the metrics.enabled config to the configmap of the agent so we can display to the user the message of agent update required for metrics in the UI

galantaa (approved):

PR #179, opened 2023-04-27
added the network-mapper dependency

shaicantor:
public: Added the network mapper dependency to be able to map service to service network connections


undera:
I can say that I really-really don't like to copy the whole chart into our repo. Do we need the namespace to be changeable that much? In otterize's values.yaml I see there is a `namespace` value to set. Isn't it sufficient?

charts/k8s-watcher/values.yaml
```diff
@@ -48,6 +48,10 @@ helm:
   enableActions: true
   volumeSizeLimit: "256Mi"
 
+network-mapper:
+  enabled: false
```
MickaelAlliel, line 52: Merging as is is fine while disabled as default.
For the next part, I'd consider before changing the default to true (if we desire that at any point) to make sure our CI tests the installation with it both enabled/disabled on existing AND new installations of the chart.

MickaelAlliel (approved):

MickaelAlliel:
> I can say that I really-really don't like to copy the whole chart into our repo. Do we need the namespace to be changeable that much? In otterize's values.yaml I see there is a `namespace` value to set. Isn't it sufficient?

I don't like it as well, but the way the otterize's chart and our chart manages namespace is different and conflicts with one another. It will require a breaking change in our helm chart to align it (to the best-practice == letting Helm manage the namespace through `--namespace` namespaced chart and not through values)

MickaelAlliel (changes requested):
My mistake - not sure why github checks did not see it but bump the `k8s-watcher/Chart.yaml` minor version and add to the readme config table the new keys you added to the values.yaml
You can use this https://www.tablesgenerator.com/markdown_tables or your IDE plugins to helm with the table format, manually is a bit of a pain

shaicantor:
> My mistake - not sure why github checks did not see it but bump the `k8s-watcher/Chart.yaml` minor version and add to the readme config table the new keys you added to the values.yaml You can use this https://www.tablesgenerator.com/markdown_tables or your IDE plugins to helm with the table format, manually is a bit of a pain

@MickaelAlliel , I see that the minor versions are bumped automatically, no?

MickaelAlliel:
> > My mistake - not sure why github checks did not see it but bump the `k8s-watcher/Chart.yaml` minor version and add to the readme config table the new keys you added to the values.yaml You can use this https://www.tablesgenerator.com/markdown_tables or your IDE plugins to helm with the table format, manually is a bit of a pain
> 
> @MickaelAlliel , I see that the minor versions are bumped automatically, no?

Patch version is bumped automatically.
I think adding a whole new dependency to the chart is a valid reason for minor bump :)

shaicantor:
> 

done

MickaelAlliel (approved):

PR #190, opened 2023-05-08
feat: pin version to v0.1.17 on public-ecr

nirbenator:
# Description
network mapper images for both deployment and daemonset were hosted on docker-hub which is causing issues to some of our customers reaching dockerhub limits.
also, the latest default tag in their chart could potentially cause breaking changes.

# manual testing
i uploaded the images to ECR and checked on production that deployment\daemonset are working and validated that we get rows in `event_networkmap_changes` @shaicantor - is there anything else i should do to check that we're still working?

### DIFF
--- a/charts/k8s-watcher/charts/network-mapper/values.yaml
+++ b/charts/k8s-watcher/charts/network-mapper/values.yaml
@@ -1,8 +1,8 @@
 # Local configuration for network mapper chart
 mapper:
-  repository: otterize
+  repository: public.ecr.aws/g8r5w0b5
   image: network-mapper
-  tag: latest
+  tag: v0.1.17
   pullPolicy:
 
   resources: { }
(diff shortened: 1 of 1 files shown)


nirbenator:
public: add network-mapper images to public ecr repository

MickaelAlliel (approved):

charts/k8s-watcher/charts/network-mapper/values.yaml
```diff
@@ -1,8 +1,8 @@
 # Local configuration for network mapper chart
 mapper:
-  repository: otterize
+  repository: public.ecr.aws/g8r5w0b5
   image: network-mapper
-  tag: latest
+  tag: v0.1.17
```
Moran-k, line 5: Is the version static?

nirbenator: i belive that we should definitly pin it until we test a newer version... having latest is a really bad practice especially wen we'er not releasing it

Moran-k: Cool, i thought we are building the image

PR #193, opened 2023-05-10
Cu-860qqrfeg daemon proxy support

michaeldeift:
public: Support proxies in daemonset

### DIFF
--- a/charts/k8s-watcher/templates/configmap-daemon.yaml
+++ b/charts/k8s-watcher/templates/configmap-daemon.yaml
@@ -19,6 +19,7 @@ data:
     [[outputs.http]]
       url = "https://app.komodor.com/metrics-collector/api/v1/collect"
       content_encoding = "gzip"
+      use_system_proxy = ${USE_PROXY}
       [outputs.http.headers]
         X-ACCOUNT-ID = "${ACCOUNT_ID}"
 
--- a/charts/k8s-watcher/templates/configmap-daemon.yaml
+++ b/charts/k8s-watcher/templates/configmap-daemon.yaml
@@ -19,7 +19,11 @@ data:
     [[outputs.http]]
       url = "https://app.komodor.com/metrics-collector/api/v1/collect"
       content_encoding = "gzip"
-      use_system_proxy = ${USE_PROXY}
+      {{- if .Values.proxy.https }}
+      http_proxy_url = "${KOMOKW_HTTPS_PROXY_URL}"
+      {{- else if .Values.proxy.http }}
+      http_proxy_url = "${KOMOKW_HTTP_PROXY_URL}"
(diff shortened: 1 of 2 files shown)


charts/k8s-watcher/templates/configmap-daemon.yaml
```diff
@@ -19,6 +19,11 @@ data:
     [[outputs.http]]
       url = "https://app.komodor.com/metrics-collector/api/v1/collect"
       content_encoding = "gzip"
+      {{- if .Values.proxy.https }}
+      http_proxy_url = "${KOMOKW_HTTPS_PROXY_URL}"
+      {{- else if .Values.proxy.http }}
+      http_proxy_url = "${KOMOKW_HTTP_PROXY_URL}"
+      {{- end }}
```
Noy-Simon, line 26: did you check it ? 
i dont think it works inside a string value

charts/k8s-watcher/templates/daemonset.yaml
```diff
@@ -48,6 +48,7 @@ spec:
                   name: {{ include "k8s-watcher.name" . }}-secret
                   key: apiKey
                   {{- end }}
+{{- include "k8s-watcher.proxy-conf" . }}
```
Noy-Simon, line 51: need an end tag no?

michaeldeift: no..it's the same in agent's file

charts/k8s-watcher/templates/configmap-daemon.yaml
michaeldeift: Yes..it works..the whole file is defined like this

Noy-Simon: even the ifs ?

michaeldeift: yes...tested with --set proxy.http="something" and --set proxy.https="sdad"

charts/k8s-watcher/templates/configmap-daemon.yaml
```diff
@@ -19,6 +19,11 @@ data:
     [[outputs.http]]
       url = "https://app.komodor.com/metrics-collector/api/v1/collect"
       content_encoding = "gzip"
+      {{- if .Values.proxy.https }}
+      http_proxy_url = "${KOMOKW_HTTPS_PROXY_URL}"
+      {{- else if .Values.proxy.http }}
+      http_proxy_url = "${KOMOKW_HTTP_PROXY_URL}"
```
Noy-Simon, line 25: why else if with the same condition?

Noy-Simon (approved):

PR #194, opened 2023-05-10
Add sidecar to monitor agent container

MickaelAlliel:
public: add sidecar container (supervisor) to monitor the komodor agent container in case of crashes - disabled by default

CU-860qt1vpe

### DIFF
--- a/charts/k8s-watcher/values.yaml
+++ b/charts/k8s-watcher/values.yaml
@@ -57,6 +57,9 @@ supervisor:
     requests:
       cpu: 0.1
       memory: 256Mi
+  servers:
+    healthCheck:
+      port: 8089
 metrics:
   enabled: false
 daemon:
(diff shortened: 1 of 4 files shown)


charts/k8s-watcher/values.yaml
```diff
@@ -48,6 +48,18 @@ helm:
...
 
+supervisor:
+  enabled: false
+  image:
+    repository: public.ecr.aws/g8r5w0b5/supervisor
+    pullPolicy: IfNotPresent
+  resources:
+    requests:
+      cpu: 0.1
+      memory: 256Mi
```
Moran-k, line 59: Do we need that much? This means we need 512MB just to start our agent

Moran-k (approved):

charts/k8s-watcher/values.yaml
MickaelAlliel: Probably need less. We will only need memory when we actually fetch data to send the crash report then release it back

MickaelAlliel: I'll tweak everything before enabling by default

PR #198, opened 2023-05-14
remove-limits-from-watcher

nirbenator:


nirbenator:
public:

charts/k8s-watcher/values.yaml
```diff
@@ -29,9 +29,6 @@ proxy:
   no_proxy: ""
 
 resources:
-  limits:
-    cpu: 1
-    memory: 4Gi
```
MickaelAlliel, line 34: Instead of removing the field completely, change to `limits: {}`
This way it will allow overriding a specific subfield `--set resources.limits.cpu=1` without getting and error tha `limits` doesnt exists

MickaelAlliel (changes requested):

PR #204, opened 2023-05-21
chore: change public ecr alias

nirbenator:


MickaelAlliel (approved):

nirbenator:
public: Changing images to use public ECR alias

PR #205, opened 2023-05-21
CU-860qwu0c2: Add node capacity metrics

YadgarYossi:
We need this data in order to summarize it, then use it to calculate the cluster scope cost, in terms of allocation.


Noy-Simon (approved):

YadgarYossi:
public: Collect node capacity metrics in addition to the rest of the existing ones.

PR #212, opened 2023-06-20
Combine network mapper

Moran-k:
## Motivation:

Integrate network mapper charts from [PR](https://github.com/komodorio/helm-charts/pull/192) 

## Changes:

* Add the network-mapper container into k8s-watcher deployment
* Add the network-mapper-sniffer into k8s-watcher daemonset 
* Move metrics stuff into a subfolder

## Clickup tasks
CU-860qvnx0u
CU-860qtrpvn
CU-860qtrpqn


charts/k8s-watcher/templates/network-mapper/serviceaccount.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.serviceAccount.create -}}
```
MickaelAlliel, line 1: This file is confusing, it's not creating a service account but creating the Bindings to the existing k8s watcher service account

charts/k8s-watcher/templates/network-mapper/cluster-role.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - apiGroups:
+      - 'apps'
+    resources:
+      - 'deployments'
+      - 'daemonsets'
+      - 'replicasets'
+      - 'statefulsets'
+    verbs:
+      - 'get'
+{{ if or (and (eq .Values.network_mapper.global.allowGetAllResources nil) .Values.network_mapper.allowGetAllResources) .Values.network_mapper.global.allowGetAllResources}}
```
MickaelAlliel, line 25: Can be simplified: (do check the syntax, i freehanded it)
```suggestion
{{ if or (((.Values.network_mapper).global).allowGetAllResources) (((.Values.network_mapper).global).allowGetAllResources}}
```

charts/k8s-watcher/templates/network-mapper/serviceaccount.yaml
Moran-k: Will rename to roles-binding.yaml

charts/k8s-watcher/templates/network-mapper/cluster-role.yaml
Moran-k: On second thought - you are right (+1)

MickaelAlliel (approved):

PR #213, opened 2023-06-21
feat: add clusterName as a required field

nirbenator:


charts/k8s-watcher/templates/configmap.yaml
```diff
@@ -7,4 +7,5 @@ metadata:
   {{- end }}
 data:
   komodor-k8s-watcher.yaml: |
-{{- include "watcher.values" . | nindent 4 }}
\ No newline at end of file
+{{- include "watcher.values" . | nindent 4 }}
+    clusterName: {{ .Values.clusterName | required "Clustername is rquired" }}
```
Moran-k, line 11: Add to readme

undera, line 11: Our `clusterName` lives under `watcher` AFAIK. Is there a mistake to use `.Values.clusterName`?

undera: Also a misprint there in `Clustername is rquired` and also `Clustername`!=`clusterName`

nirbenator: whops for both , good catch :)

nirbenator:
public: watcher.clusterName is now a required value

PR #214, opened 2023-06-25
CU-85ztcd9pw - fix missing clustername in generate kube templates

Moran-k:
* Add to generate-kube the cluster name
* Update the readme file.


Moran-k:
non-public: fix missing clusterName in generate-kube

nirbenator (approved):

PR #225, opened 2023-07-24
Add namespace to network-mapper cluster-role.yaml

Moran-k:


Moran-k:
public: Use namespace in network-mapper

PR #229, opened 2023-08-10
CU-860rdgtu0 - separate metrics inventory plugin into deploy

Noy-Simon:
public: improve metric collection performance

### DIFF
--- a/charts/k8s-watcher/templates/deployment.yaml
+++ b/charts/k8s-watcher/templates/deployment.yaml
@@ -41,9 +41,7 @@ spec:
         {{- include "metrics.init" . | indent 8 }}
       {{- end }}
       containers:
-        {{- if .Values.metrics.enabled }}
         {{- include "metrics.container" .        | indent 8 }}
-        {{- end }}
         {{- include "network_mapper.deployment.container" . | nindent 8 }}
         
 {{- if ((.Values.supervisor)).enabled }}
@@ -161,6 +159,7 @@ spec:
             failureThreshold: 3
             successThreshold: 1
       volumes:
+        {{- include "metrics.deploy.volumes" . | nindent 8 }}
         - name: configuration
           configMap:
             name: {{ include "k8s-watcher.name" . }}-config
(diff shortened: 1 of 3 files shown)


charts/k8s-watcher/templates/metrics/deploy.yaml
```diff
@@ -0,0 +1,43 @@
...
+  {{- end }}
+  {{- if .Values.watcher.clusterName }}
+  - name: CLUSTER_NAME
+    value: {{ .Values.watcher.clusterName }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
+{{- define "metrics.daemonset.volumes" }}
+{{- if and .Values.metrics (.Values.metrics.enabled | default false) }}
```
MickaelAlliel, line 32: wont this volume definition clash with the one the daemonset uses?

charts/k8s-watcher/templates/metrics/deploy.yaml
```diff
@@ -28,7 +28,7 @@
 {{- end }}
 {{- end }}
 
-{{- define "metrics.daemonset.volumes" }}
+{{- define "metrics.deploy.volumes" }}
 {{- if and .Values.metrics (.Values.metrics.enabled | default false) }}
 - name: {{ include "k8s-watcher.name" . }}-metrics-config
```
MickaelAlliel, line 33: The volume name as well - it is the same of the daemonset

charts/k8s-watcher/templates/deployment.yaml
```diff
@@ -41,9 +41,7 @@ spec:
         {{- include "metrics.init" . | indent 8 }}
       {{- end }}
       containers:
-        {{- if .Values.metrics.enabled }}
         {{- include "metrics.container" .        | indent 8 }}
```
MickaelAlliel, line 44: why did you remove the condition of if metrics is enabled?

charts/k8s-watcher/templates/metrics/deploy.yaml
Noy-Simon: nope in the daemon its -daemon-config

charts/k8s-watcher/templates/deployment.yaml
Noy-Simon: its already part of the metrics.container 
its redundant to do it twice

MickaelAlliel (approved):

PR #235, opened 2023-08-14
CU-85ztt27jc - Add test to ci 

Noy-Simon:


MickaelAlliel (commented):
If we're only checking the helm-chart -- We probably want to only run this in branch and not master.
You can make it run only on branches in the if condition `&& branch != master` or something like that

MickaelAlliel (approved):

PR #236, opened 2023-08-17
CU-85zttjvr1 - make api server url configurable

Noy-Simon:
public: add api server url configuration to values

### DIFF
--- a/charts/k8s-watcher/values.yaml
+++ b/charts/k8s-watcher/values.yaml
@@ -86,7 +86,8 @@ daemon:
     requests:
       cpu: 0.1
       memory: 256Mi
-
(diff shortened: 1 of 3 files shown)


charts/k8s-watcher/values.yaml
```diff
@@ -86,6 +86,8 @@ daemon:
     requests:
       cpu: 0.1
       memory: 256Mi
+  configuration:
+    apiServerUrl: "https://kubernetes.default.svc"
```
MickaelAlliel, line 90: Where has the `cluster.local` gone to?

MickaelAlliel (approved):

PR #237, opened 2023-08-17
Add support for configuring Tags on the agent

MickaelAlliel:
CU-860rd7x06

public: support `tags` and `watcher.tags` configuration as agent labeling


charts/k8s-watcher/README.md
```diff
@@ -94,115 +94,116 @@ To install the chart directly with kubectl, use the manifests located in `./kube
...
+| `apiKey`                                                   | Komodor kubernetes api key (required if `existingSecret` not specified)                                                                                                                                                                                   | ``                                  |
+| `existingSecret`                                           | Existing kubernetes secret resource containing Komodor kubernetes apiKey (required if `apiKey` not specified)                                                                                                                                             | ``                                  |
+| `tags`                                                     | Tags the agent in order to identify it based on `key:value` properties separated by semicolon (`;`) - example: `--set tags="env:staging;team:product-a"` --- Can also be set in the values under `watcher.tags` as a dictionary of key:value strings      | ``                                  |
```
shaicantor, line 101: maybe add to the description
```Tags the agent in order to identify it and the cluster based on....```

Comment (MickaelAlliel) [reply charts/k8s-watcher/README.md:101]: Ehmm.. that's exactly what is there already? :)

File: charts/k8s-watcher/templates/deployment.yaml [hunk 2]
```diff
@@ -122,6 +122,10 @@ spec:
 {{- else }}
                   name: {{ include "k8s-watcher.name" . }}-secret
                   key: apiKey
+{{- end }}
+{{- if not (hasKey .Values.watcher "tags") }}
+            - name: KOMOKW_TAGS
+              value: {{ .Values.tags | default "" | quote }}
```
Moran-k, line 128: Just to make sure this scenario is covered.
where or when do we set watcher.tags?
because in case it is not set you assume we have .Values.tags and it is possible we dont.

Moran-k: My bad, i see the default is empty.

Moran-k (approved):

charts/k8s-watcher/templates/deployment.yaml
MickaelAlliel: It's just that its easier to set during installation a string `env:prod;team:a` rather than setting a yaml list in command line.
That's why I prefered to use this way - but still want to preserve the possibility of using a `-f values.yaml` in case customers want to use that

MickaelAlliel: So yes -- default is empty, and we recommend setting `--set tags` rather than `--set watcher.tags`

PR #242, opened 2023-08-24
CU-85ztv4pr8 - Fix network mapper cannot be disabled

Moran-k:
Fix template if condition that enables/disables the network-mapper


Moran-k:
public: Fix an issue that prevented disabling network mapper or metrics features

nirbenator (approved):

PR #243, opened 2023-08-24
[skip ci] CU-85ztv4xmu - fix pipeline not blocking before publishing the chart

Moran-k:


Moran-k:
non-public: fix pipeline

nirbenator (approved):

PR #245, opened 2023-08-27
CU-85ztvkk2y - Update values.yaml with Update values with logNamespacesAllowList

ElisarEisenbach:


nirsht (approved):

ElisarEisenbach:
non-public: Update values.yaml with Update values with logNamespacesAllowList

PR #254, opened 2023-09-18
CU-860rr4w3 - image pull secrets

nirbenator:


Moran-k:
what we have in the service account is not enough?

nirbenator:
much better! closing this pr

PR #259, opened 2023-10-11
Komodor agent vlun scan

Moran-k:
Scan all images in `komodor-agent` values.yaml file using mend.

The scan script:
* scan for critical vulnerabilities only
* Skip critical vulnerabilities without a fixed version
* Ignore vulnerabilities that are in the `exclude.yaml`

In case a critical vulnerability is found, the build will fail.
Build failure notification will be sent to `#critical-security-vulns` slack channel


nirbenator (approved):
king! the tests! 🤤

nirbenator (approved):

PR #261, opened 2023-10-15
feat(metrics): Improve metrics inventory performance and support large scale clusters

Noy-Simon:
public: Improve metrics inventory performance and support large scale clusters


charts/komodor-agent/values.yaml
nirbenator, line 1: @Noy-Simon - if we're removing metrics from the komodor-agent (or the deployment), id move metrics values  from components.komodorAgent.metrics to componets.komodorDaeon.metrics

Noy-Simon: @nirbenator  done

charts/komodor-agent/templates/metrics/configmap-daemon.yaml
```diff
@@ -33,5 +33,20 @@ data:
       namedrop = ["kubernetes_system_container"]
       [inputs.kubernetes.tags]
         measure_type = "usage"
+        
+    [[inputs.kube_inventory]]
+      ## URL for the Kubernetes API
+      url = {{ .Values.communications.apiServerUrl | quote }}
+      url_kubelet = "https://$NODE_IP:10250"
```
Moran-k, line 40: Is this port always 10250? Don't we need it to be configurable?

Moran-k (approved):
Looks good to me

charts/komodor-agent/templates/metrics/configmap-daemon.yaml
galHalup: it's the kubelet port, I guess it won't be changed soon

PR #262, opened 2023-10-16
CU-865d83p3u - First full release of komodor-agent

Moran-k:
1. fix the missing namespace in one pipeline step
2. Verify full release flow


Moran-k:
public: Add a new `komodor-agent` chart that will replace the current `k8s-watcher` chart

MickaelAlliel (approved):

PR #270, opened 2023-11-09
Helm uninstall not working due to lack permissions

Moran-k:
Enable secrets by default when helm capability is enabled.
Allow deleting all resources when helm capability is enabled (Allow * delete in cluster role). 
Do not bump k8s-watcher chart version if not needed.


Moran-k:
public: Fix helm uninstall request is not working due to missing permissions.

nirbenator (approved):

PR #271, opened 2023-11-14
Changes after CS review - CU-860rzrtfh

Moran-k:
* Move communications  to the bottom of the values file
* Add `--create-namespace`  to the migrations script


Moran-k:
none-public: reorder values file

MickaelAlliel (approved):

PR #274, opened 2023-11-27
Add node-enricher to komodor-agent chart

rannyeli:
public: Install node-enricher service as part of the komodor-agent daemonset

### DIFF
--- a/charts/komodor-agent/templates/node-enricher/_containers.tpl
+++ b/charts/komodor-agent/templates/node-enricher/_containers.tpl
@@ -1,0 +1,24 @@
+{{- define "node_enricher.daemonset.container" }}
+{{- if .Values.capabilities.nodeEnricher }}
+- name: node-enricher
+  image: {{ .Values.imageRepo }}/{{ .Values.components.komodorDaemon.nodeEnricher.image.name}}:{{ .Values.components.komodorDaemon.nodeEnricher.image.tag }}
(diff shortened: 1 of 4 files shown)


rannyeli:
Task linked: [CU-860t053vx [Agent] Implement a new event listener for spot interruption events](https://app.clickup.com/t/860t053vx)

charts/komodor-agent/values.yaml
```diff
@@ -54,6 +54,8 @@ capabilities:
   metrics: true
   # capabilities.networkMapper -- (bool) Enable network mapping capabilities by the komodor agent
   networkMapper: true
+  # capabilities.nodeEnricher -- (bool) Enable node enricher capabilities by the komodor agent
+  nodeEnricher: true
```
MickaelAlliel, line 58: I'd start by disabling it by default, testing on our own clusters for a few days and only then enabling it for customers. We do have the new agent release process which should validate it enough before the actual release but sync with @Moran-k to make sure how it works.

charts/komodor-agent/templates/node-enricher/_containers.tpl
```diff
@@ -0,0 +1,21 @@
+{{- define "node_enricher.daemonset.container" }}
+{{- if .Values.capabilities.nodeEnricher }}
```
MickaelAlliel, line 2: No probes?

MickaelAlliel:
Run `make generate-readme` :)

PR #275, opened 2023-11-28
CU-867981mn2 - Bugfixes from POC inputs

Moran-k:
Use different node affinity between deployment and daemonset
Remove resources from the metrics daemon


Moran-k:
public: 
* Add node affinity to agent daemon
* Remove metrics daemon resources to align with old chart

nirbenator (approved):

PR #276, opened 2023-11-28
CU-867985xym - Generate GitHub release from GH issue messages

Moran-k:
Generate GitHub release for the RC version as `pre-release`

The new script looks for `public: ` messages in the repo between the last GA version and the current RC version.
In addition, it also looks for all `public: ` messages in the `komodor-agent` repo between the last and (last - 1) tag and adds the message to the release.


nirbenator (approved):
Looking great, do we have the check enabled in the warden to enforce public?

Moran-k:
public: Auto-generate pre-release

PR #277, opened 2023-11-29
CU-867986mhv - Remove limits from metrics daemon

Moran-k:
Remove limits from metrics daemon,
same as we have in k8s-watcher chart


Moran-k:
public: Remove limit from metrics daemon

nirbenator (approved):

PR #278, opened 2023-11-30
CU-86798y3bd - set resources for metrics container

Moran-k:
We checked the P95 from timestream for all customer and decided on suitable values.


galHalup (approved):

nirbenator (approved):

Moran-k:
public: set requests and limits resources for metrics container

PR #290, opened 2024-01-01
cfg: add mgmtServerHost communications configuration

MickaelAlliel:
Related to: https://github.com/komodorio/komodor-agent/pull/335

CU-860qv9ngg

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -256,8 +256,8 @@ steps:
       - helm repo add komodorio https://helm-charts.komodor.io
       - helm repo update
       - helm get values komodor-agent-staging -n komodor-staging > staging-values-to-reuse.yaml
+      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -n komodor-staging -f staging-values-to-reuse -f ./.buildkite/pipeline_scripts/staging-values.yaml --dry-run
       - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values-to-reuse.yaml -f ./.buildkite/pipeline_scripts/staging-values.yaml --namespace=komodor-staging --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY
-      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -n komodor-staging -f ./.buildkite/pipeline_scripts/staging-values.yaml --dry-run
 
       - ./.buildkite/pipeline_scripts/update_agent.sh staging komodor-staging komodor-agent-production $$STAGING_PRODUCTION_API_KEY
     agents:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -256,8 +256,8 @@ steps:
       - helm repo add komodorio https://helm-charts.komodor.io
       - helm repo update
       - helm get values komodor-agent-staging -n komodor-staging > staging-values.yaml
+      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values.yaml --set clusterName=komodor-staging-eks --namespace=komodor-staging --set imagePullSecret=docker-cfg-komodorio --set communications.serverHost=https://staging.app.komodor.com --set communications.tasksServerHost=https://staging.app.komodor.com --set communications.wsHost=wss://staging.app.komodor.com --set communications.mgmtServerHost=https://staging.app.komodor.com --set capabilities.telemetry.enable=false --set communications.telemetryServerHost=https://staging.telemetry.komodor.com --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY --set allowedResources.secret=true --set capabilities.event.redact="{.*KEY.*,.*key.*,.*BUGSNAG.*}"
       - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -n komodor-staging -f staging-values.yaml --dry-run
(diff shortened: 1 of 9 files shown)


.buildkite/pipeline.yml
```diff
@@ -256,8 +256,8 @@ steps:
       - helm repo add komodorio https://helm-charts.komodor.io
       - helm repo update
       - helm get values komodor-agent-staging -n komodor-staging > staging-values.yaml
+      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values.yaml --set clusterName=komodor-staging-eks --namespace=komodor-staging --set imagePullSecret=docker-cfg-komodorio --set communications.serverHost=https://staging.app.komodor.com --set communications.tasksServerHost=https://staging.app.komodor.com --set communications.wsHost=wss://staging.app.komodor.com --set communications.mgmtServerHost=https://staging.app.komodor.com --set capabilities.telemetry.enable=false --set communications.telemetryServerHost=https://staging.telemetry.komodor.com --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY --set allowedResources.secret=true --set capabilities.event.redact="{.*KEY.*,.*key.*,.*BUGSNAG.*}"
```
nirbenator, line 259: @Moran-k 
whats the point of passing values this way? if we already have a `staging-values` file?

MickaelAlliel: I was about to ask @Moran-k if we should delete the `staging-values.yaml` file in the CI folder.. See line 258 here, we generate it, and I assume it overwrites what we have in the folder, so I can only think it was some kind of test we committed by mistake

MickaelAlliel: Also I do think we should have files instead and do `-f current-values-from-get.yaml -f staging-values.yaml` to make it cleaner

nirbenator: i see we generate it, but why do we do it ? we should just have values file per env

MickaelAlliel: the generated ones is our way of doing `--reuse-values` without the downside of reuse-values

nirbenator: i know i know, but instead of running --reuse-values we can quite literally reuse the saved `values` file. anyways. lets ask @Moran-k  if there's any reason to do so and if not i would just delete it

Moran-k: `staging-values.yaml` was not supposed to be committed,
As mick said the `get values` simulates the reuse same as we show to our customers now, so i did the same,
and i added all the parameters to the command to be safe if one day someone uninstalls the agent from the cluster or if we will recreate the cluster....

After should thinking, maybe it is better and clearer to remove the get values + all the `--set` and use a value file....

nirbenator: i vote for using `-f staging-values` and removing `--set`s unless this is the only place where we are doing the get values update values trick (which is good to test somewhere)

charts/komodor-agent/values.yaml
```diff
@@ -343,3 +343,4 @@ communications:
   wsHost: wss://app.komodor.com
   tasksServerHost: *serverHost
   telemetryServerHost: https://telemetry.komodor.com
+  mgmtServerHost: *serverHost
```
Moran-k, line 346: I don't see we use this value in the template....

nirbenator: we copy everything `communications` to config,yaml

MickaelAlliel: Yes, thanks @nirbenator

.buildkite/pipeline.yml
```diff
@@ -255,9 +255,9 @@ steps:
...
-      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values.yaml --set clusterName=komodor-staging-eks --namespace=komodor-staging --set imagePullSecret=docker-cfg-komodorio --set communications.serverHost=https://staging.app.komodor.com --set communications.tasksServerHost=https://staging.app.komodor.com --set communications.wsHost=wss://staging.app.komodor.com --set capabilities.telemetry.enable=false --set communications.telemetryServerHost=https://staging.telemetry.komodor.com --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY --set allowedResources.secret=true --set capabilities.event.redact="{.*KEY.*,.*key.*,.*BUGSNAG.*}"
+      - helm get values komodor-agent-staging -n komodor-staging > staging-values-to-reuse.yaml
+      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values-to-reuse.yaml -f ./.buildkite/pipeline_scripts/staging-values.yaml --namespace=komodor-staging --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY
+      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -n komodor-staging -f ./.buildkite/pipeline_scripts/staging-values.yaml --dry-run
```
Moran-k, line 260:
```suggestion
      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -n komodor-staging -f staging-values-to-reuse.yaml -f ./.buildkite/pipeline_scripts/staging-values.yaml --dry-run
      - helm upgrade --install komodor-agent-staging komodorio/komodor-agent -f staging-values-to-reuse.yaml -f ./.buildkite/pipeline_scripts/staging-values.yaml --namespace=komodor-staging --set apiKey=$$KOMODOR_AGENT_STAGING_API_KEY
      
```

The dry run should be before the actual upgrade, and should also include the reuse-values file

Moran-k (approved):
LGTM

Moran-k:
@MickaelAlliel  - merge from master please

.buildkite/tests/values_capabilities_proxy_test.py
```diff
@@ -1,19 +1,21 @@
-import pytest
 import time
+
+import pytest
+
 from config import NAMESPACE
-from fixtures import setup_cluster, kube_client, cleanup_agent_from_cluster
```
Moran-k, line 4: Guessing this was removed by mistake... (it is required)

.buildkite/tests/values_capabilities_test.py
```diff
@@ -1,12 +1,12 @@
 import time
+
+import yaml
+
 from config import BE_BASE_URL, NAMESPACE, RELEASE_NAME
-from helpers.utils import get_filename_as_cluster_name
-from fixtures import setup_cluster, cleanup_agent_from_cluster
```
Moran-k, line 4: same

.buildkite/tests/values_capabilities_proxy_test.py
MickaelAlliel: Oops.. linter got the best of me

PR #292, opened 2024-01-11
added restarts total to metrics collector inventory plugin

erez-rabih:
we need the restarts total for each container to support the reliability check
cantor found out this field exists in the kube inventory plugin
image

we tested on staging - changed the configmap and restarted the daemon set  -> we saw restarts_count is added to the staging timestream DB


erez-rabih:
cu-86bx176r8

Moran-k (approved):

PR #313, opened 2024-04-10
Cu 86by9b9xp support gather interval

michaeldeift:
Update telegraf image to a version supporting poll_interval
Add poll_interval to configmap

### DIFF
--- a/charts/komodor-agent/templates/metrics/configmap-daemon-windows.yaml
+++ b/charts/komodor-agent/templates/metrics/configmap-daemon-windows.yaml
@@ -38,6 +38,7 @@ data:
       url = {{ .Values.communications.apiServerUrl | quote }}
       url_kubelet = "https://$NODE_IP:10250"
       namespace = ""
+      poll_interval = "${INTERVAL}"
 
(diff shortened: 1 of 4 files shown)


charts/komodor-agent/templates/metrics/configmap-daemon-windows.yaml
```diff
@@ -38,6 +38,7 @@ data:
       url = {{ .Values.communications.apiServerUrl | quote }}
       url_kubelet = "https://$NODE_IP:10250"
       namespace = ""
+      poll_interval = "${INTERVAL}"
```
Moran-k, line 41: where is this `INTERVAL` parameter set?

Moran-k (approved):

charts/komodor-agent/templates/metrics/configmap-daemon-windows.yaml
michaeldeift: It is set by agents-service in ResponseTelegrafIdentifyClusterConfiguration struct

PR #325, opened 2024-06-30
Fix(komodor-agent): Wrong values

Moran-k:
* Fix the wrong container name when enabling `enableMemLimitChecks`
* Fix the wrong value check if Helm is enabled.
* Update CI `legacy k8s versions` test to check for k8s-version 1.27


Moran-k:
CU-86bzcmk4w

nirbenator (approved):

PR #331, opened 2024-09-05
CU-86c081bq5 - Fix formatting of Tags when passed as helm values as string or as map

MickaelAlliel:
## Changelog

* Pass `tags` configuration in the config file instead of environment variable
* Fix parsing of tags when provided as a string with semicolon delimiter

### DIFF
--- a/charts/komodor-agent/templates/configmap.yaml
+++ b/charts/komodor-agent/templates/configmap.yaml
@@ -39,5 +39,21 @@ data:
     resources:
       {{- toYaml .Values.allowedResources | trim | nindent 6 }}
 
+    {{- if .Values.tags }}
+    {{- if and .Values.tags (kindIs "map" .Values.tags) }}
+    tags: {{ .Values.tags | nindent 6 }}
+    {{- else if and .Values.tags (kindIs "string" .Values.tags) }}
+    {{ $tags := .Values.tags | splitList ";" }}
+    {{ $tagMap := dict }}
+    {{ range $tags }}
+      {{ $pair := splitList ":" . }}
+      {{ $key := index $pair 0 }}
+      {{ $value := index $pair 1 }}
+      {{ $_ := set $tagMap $key $value }}
+    {{ end }}
+    tags: {{ $tagMap | toYaml | nindent 6 }}
+    {{- end }}
+    {{- end }}
+
   installed-values.yaml: |
     {{ toYaml .Values | nindent 4 }}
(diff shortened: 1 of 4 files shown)


nirbenator:
Task linked: [CU-86c081bq5 Fix formatting of Tags when passed as helm values as string or as map](https://app.clickup.com/t/86c081bq5)

charts/komodor-agent/templates/configmap.yaml
```diff
@@ -39,5 +39,21 @@ data:
...
+    {{- if and .Values.tags (kindIs "map" .Values.tags) }}
+    tags: {{ .Values.tags | nindent 6 }}
+    {{- else if and .Values.tags (kindIs "string" .Values.tags) }}
+    {{ $tags := .Values.tags | splitList ";" }}
+    {{ $tagMap := dict }}
+    {{ range $tags }}
+      {{ $pair := splitList ":" . }}
+      {{ $key := index $pair 0 }}
+      {{ $value := index $pair 1 }}
+      {{ $_ := set $tagMap $key $value }}
```
Moran-k, line 52: Maybe add a check to ensure the string contains both a key and a value 
```suggestion
      {{- if eq (len $pair) 2 }}
        {{ $key := index $pair 0 }}
        {{ $value := index $pair 1 }}
        {{ $_ := set $tagMap $key $value }}
      {{- end }}
```

charts/komodor-agent/templates/configmap.yaml
```diff
@@ -39,5 +39,21 @@ data:
...
+    {{ end }}
+    tags: {{ $tagMap | toYaml | nindent 6 }}
+    {{- end }}
```
Moran-k, line 55: low chances, but maybe exist with error in case the provided type is unexpected
```suggestion
    {{- else }}
        {{ fail "Invalid type for .Values.tags. Expected map or string." }}
    {{- end }}
```

charts/komodor-agent/values.yaml
```diff
@@ -6,7 +6,7 @@ apiKeySecret:
 createNamespace: true
 # tags -- (dict) Tags the agent in order to identify it based on `key:value` properties separated by semicolon (`;`)
 # example: `--set tags="env:staging;team:product-a"` --- Can also be set in the values under `tags` as a dictionary of key:value strings
```
Moran-k, line 8: Maybe update the docstring to use yaml dict instead of the string?

Moran-k (commented):
minor improvement suggestions, nonblocking

Moran-k (approved):
LGTM

PR #334, opened 2024-09-24
CU-86c0fwhr0 - feat: allow not creating rbac at all

nirbenator:
# Motivation
Some accounts have security policies that require teams to validate changes in Kubernetes RBAC. By not creating RBAC, the Helm chart will avoid generating any cluster roles or bindings. However, modifying RBAC places full responsibility on the account for managing permission errors, which can lead to unexpected behavior. This approach is not recommended and should be used with caution.


nirbenator:
Task linked: [CU-86c0fwhr0 feat: allow not creating rbac at all](https://app.clickup.com/t/86c0fwhr0)

Moran-k (approved):

PR #336, opened 2024-11-03
Add toleration to ci values

Moran-k:
Allow komodor agent to be scheduled on nodes with `komodor.io/sensitive` taint


Moran-k:
CU-86c0wh5fk

PR #338, opened 2024-11-04
[CU-86c0x1d85] Add missing toleration on daemon

galHalup:
This was in the past and probably was missed when we moved to the komodor-agent from the watcher

https://github.com/komodorio/helm-charts/blame/b3ea76976400b58fd4df1f513b1ea5fd4d5127d0/charts/k8s-watcher/values.yaml#L72


nirbenator:
Task linked: [CU-86c0x1d85 daemon is not installed in customer env - add tolerations to prevent it from happening](https://app.clickup.com/t/86c0x1d85)

Moran-k (approved):

PR #345, opened 2024-12-22
CU-86c1djzn9 - fix(ci): Update tests after removing installationId authentication type from BE endpoints 

Moran-k:
Fix CI to use relevant test API keys

### DIFF
--- a/.buildkite/tests/values_capabilities_test.py
+++ b/.buildkite/tests/values_capabilities_test.py
@@ -20,7 +20,7 @@ def wait_for_metrics(container_name, pod_name):
                f"clusterName={CLUSTER_NAME}&namespace={NAMESPACE}&podName={pod_name}&"
                f"fromEpoch={start_time}&pageSize=1&timeWindow=1d&endTime={end_time}&"
                f"aggregationTypes=p96&aggregationTypes=p99")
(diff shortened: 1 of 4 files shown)


.buildkite/tests/values_capabilities_test.py
```diff
@@ -20,7 +20,7 @@ def wait_for_metrics(container_name, pod_name):
                f"clusterName={CLUSTER_NAME}&namespace={NAMESPACE}&podName={pod_name}&"
                f"fromEpoch={start_time}&pageSize=1&timeWindow=1d&endTime={end_time}&"
                f"aggregationTypes=p96&aggregationTypes=p99")
-        response = query_backend(url)
+        response = query_backend(url, agent_api_key=False)
```
nirbenator, line 23: i see that we also `query_backend` in `test_network_mapper`
why dont we change to agent_api_key_false there?

nirbenator (approved):
lgtm - just a small question

.buildkite/tests/values_capabilities_test.py
Moran-k: I changed the default to false, so basically, we are not using agent-api anymore

PR #355, opened 2025-02-24
adding support for setting up the secretKey in both user provided sec…

amit9192:
Mend uses external secrets and would like to utilize different key names per cluster to support easier automation of cluster onboarding.
I'm debating whether this should also be part of the Komodor agent chart managed secret but implemented it for both currently.


charts/komodor-agent/templates/watcher/_containers.tpl
```diff
@@ -66,7 +66,7 @@
       port: http-healthz
     initialDelaySeconds: 5
     periodSeconds: 5
-    failureThreshold: 3
+    failureThreshold: 3git
```
Moran-k, line 69:
```suggestion
    failureThreshold: 3
```

PR #359, opened 2025-03-03
wip

nirbenator:


charts/komodor-agent/values.yaml
```diff
@@ -149,6 +151,10 @@ components:
     securityContext: { }
     # components.komodorAgent.strategy -- Set the rolling update strategy for the komodor agent deployment
     strategy: { }
+    # components.komodorAgent.hostNetwork -- Set host network for the komodor agent
+    hostNetwork: false
+    # components.komodorAgent.dnsPolicy -- Set dns policy for the komodor agent
+    dnsPolicy: ClusterFirst
```
Moran-k, line 157: Should be  also added to `components.komodorMetrics` ?

charts/komodor-agent/values.yaml
```diff
@@ -305,6 +311,10 @@ components:
         requests:
           cpu: 0.1
           memory: 384Mi
+      # components.komodorDaemon.metrics.ports -- Override the komodor agent metrics ports configuration
+      ports:
+        # components.komodorDaemon.metrics.ports.healthCheck -- Override the health check port of the komodor agent metrics
+        healthCheck: 8090
```
Moran-k, line 317: Why not add this to all components? A customer may want to change both the metrics and the agent since another product is using the same port

Moran-k (commented):
- Update PR title and description
- A comment about network/port stuff,

PR #363, opened 2025-03-20
fix: remove hostNetwork parameter from komodorDaemonWindows configuration - not supported on windows

galHalup:
This pull request includes changes to the `komodor-agent` Helm chart to remove the `hostNetwork` configuration for the Windows daemon since it's not supported on windows.


galHalup:
CU-86c2h30e8

Moran-k (approved):

PR #391, opened 2025-06-10
separate the admission controller cluster role and service account to its own resources

MickaelAlliel:
CU-86c3ybjrg

Installed & tested locally on Kind


nirbenator (approved):
amazing
imo - we dont have to expose all this configuration to values file. just leave it as capabilities admission controller. but i think its a matter or preferences

MickaelAlliel:
> amazing imo - we dont have to expose all this configuration to values file. just leave it as capabilities admission controller. but i think its a matter or preferences

Ehh I mean we have added these options to the agent because customers asked for it, might as well be configurable :)

PR #392, opened 2025-06-11
small lefovers from migrations

Noy-Simon:


charts/komodor-agent/values.yaml
```diff
@@ -474,8 +474,6 @@ components:
       extraEnvVars: []
       # components.komodorDaemon.metrics.quiet -- Set the quiet mode for the komodor agent metrics
       quiet: false
-      # components.komodorDaemon.metrics.dcgm -- Enable collecting NVIDIA GPU metrics via DCGM
-      dcgm: true
```
Moran-k, line 478: Is this intentional?

PR #407, opened 2025-07-03
CU-86c4ac72z - Add service for otel-collector to be reacheable from other komodor pods and fix container command

MickaelAlliel:
Will allow us to route telemetry from the agent to this sidecar instead of directly to our backend
Also remove the redundant command override which caused a subprocessed to spawn


Moran-k (approved):
LGTM,
In feauture k8s versions we might want to use [PreferSameNode](https://kubernetes.io/docs/concepts/services-networking/service/#:~:text=the%20intended%20semantics.-,PreferSameNode,-Indicates%20a%20preference) which is Alpha feature in k8s 1.33

PR #411, opened 2025-07-21
add a checksum to automatically refresh controller server with new certificates and configuration if they change

MickaelAlliel:
CU-86c4md27r


Moran-k (approved):

Moran-k (commented):
Seems like it's generating a malformed annotation  value

MickaelAlliel:
> Seems like it's generating a malformed annotation value

yeah the two slashes.. helm helm helm

PR #416, opened 2025-08-11
Add helm value and env var to skip tls verification

michaeldeift:
Support setting skip TLS verification during helm installation.
This option is already supported in agent code:
	if os.Getenv("SKIP_TLS_VERIFY") != "" {
		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	}

Question - do we want to use the same http client for both k8s API and general http requests OR
have a dedicated client for general http requests with its own config vars?

### DIFF
--- a/charts/komodor-agent/values.yaml
+++ b/charts/komodor-agent/values.yaml
@@ -12,6 +12,9 @@ clusterName:
 # createRbac -- Creates the necessary RBAC resources for the agent - use with caution!
 createRbac: true
 
+# skipTlsVerify -- (bool) Skip TLS certificate verification for the agent (sets SKIP_TLS_VERIFY environment variable)
+skipTlsVerify: false
+
 # telegrafImageVersion -- (string) Telegraf version to be used
 telegrafImageVersion: &telegrafVersion v2.0.3-alpine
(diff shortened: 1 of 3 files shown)


charts/komodor-agent/values.yaml
```diff
@@ -12,6 +12,11 @@ clusterName:
 # createRbac -- Creates the necessary RBAC resources for the agent - use with caution!
 createRbac: true
 
+# skipTlsVerify -- (bool) Skip TLS certificate verification for the agent (sets SKIP_TLS_VERIFY environment variable)
+skipTlsVerify: false
+# httpRequestsSkipTlsVerify -- (bool) Skip TLS certificate verification for HTTP requests (sets HTTP_REQUESTS_SKIP_TLS_VERIFY environment variable)
+httpRequestsSkipTlsVerify: false
```
MickaelAlliel, line 18: This should most likely be nested under capabilities.tasks.httpRequests ?

michaeldeift: ok..will move it

MickaelAlliel (approved):

PR #483, opened 2025-12-10
CU-86c712czc - disable status labels and events by default for podmotion

MickaelAlliel:


MickaelAlliel:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86c712czc](https://app.clickup.com/t/86c712czc)

Moran-k (approved):

PR #560, opened 2026-04-15
feat: add UPDATE operation to HPA mutating webhook

OrBin:
## What
Add the `UPDATE` operation to the HPA mutating webhook configuration, alongside the existing `CREATE` operation.

## Why
The mutating webhook for HorizontalPodAutoscalers was only intercepting `CREATE` operations. This meant that updates to existing HPAs were not being processed by the admission controller. Adding `UPDATE` ensures the webhook also handles modifications to HPA resources.


OrBin:
CU-86c9az850

maxdbn (approved):

PR #581, opened 2026-05-21
CU-86c9x4pre - Bug: add global security context fallbacks

davidcollom:
## What changed

- Added chart-level helpers for pod and container security contexts.
- Added `global.securityContext` as a chart-wide container-level fallback.
- Added `global.podSecurityContext` as a chart-wide pod-level fallback.
- Preserved component and container-specific overrides so a workload can opt out of the global default when it needs a different security posture.
- Updated security context resolution to merge values instead of replacing them:
  - Container precedence is chart default -> global -> container/component override.
  - Pod precedence is chart default -> global -> deprecated component value -> pod override.
- Wired the fallback behavior through the Komodor agent, metrics deployment, daemonset, Windows daemonset, admission controller, kubectl proxy, GPU access, OpenTelemetry, and custom CA init paths.
- Extended pytest coverage for deprecated pod-level fallback behavior, explicit `podSecurityContext`, global pod context fallback, rendered container security contexts, and the regression where `global.securityContext` must not remove chart-provided defaults.

## Why this matters

Security context configuration was previously spread across multiple chart-specific helpers and not all containers had the same setup path. That made it harder for users to apply a consistent baseline across the full Komodor deployment, especially during first-time onboarding into clusters with Pod Security Admission, OPA/Gatekeeper, Kyverno, or organization-specific hardening policies.

A global fallback lets users define one default container security policy once, for example `runAsNonRoot`, `allowPrivilegeEscalation: false`, dropped capabilities, or a default seccomp profile, and have it apply across all chart-managed containers. This reduces the amount of values YAML users need to write and lowers the chance that one sidecar or init container is missed.

The merge behavior is important because chart-provided defaults are part of the runtime contract for some containers. Setting one global value must not strip hardened defaults from watcher or supervisor, and it must not remove required container-specific settings such as `privileged: true` for the GPU host-access container.

The override behavior is equally important because not every container has identical runtime requirements. Component-level pod contexts and container-specific security contexts still take precedence over the global defaults. That means users can set a secure baseline globally while preserving targeted exceptions for containers that require a different UID, root access, privileged mode, or platform-specific Windows behavior.

This should improve onboarding for security-conscious users because they can start with a small global policy and then adjust only the exceptional containers, instead of discovering every container-specific values path manually.

## Validation

```bash
. .venv/bin/activate ; pytest values_components_test.py -k preserves_chart_defaults
```

Result: `3 passed`.

```bash
. .venv/bin/activate ; pytest values_components_test.py
```

Result: `75 passed`.

```bash
helm lint . \
  --set apiKey=test-api-key \
  --set clusterName=test-cluster-name \
  --set global.securityContext.runAsNonRoot=true \
  --set global.podSecurityContext.runAsNonRoot=true
```

Result: `1 chart(s) linted, 0 chart(s) failed`.

```bash
helm template komodor-agent-test . \
  --namespace test-namespace \
  --set apiKey=test-api-key \
  --set clusterName=test-cluster-name \
  --set global.securityContext.runAsNonRoot=true \
  --set components.gpuAccess.enabled=true
```

Result: rendered successfully and confirmed the key regression scenario:

- watcher and supervisor keep hardened chart defaults while inheriting `runAsNonRoot: true`.
- GPU host-access keeps `privileged: true` while inheriting `runAsNonRoot: true`.


charts/komodor-agent/templates/_security.tpl
Moran-k, line 1: The new `global.securityContext` silently strips chart-provided defaults. I think this is a critical regression.

the new helper uses **replace** precedence  (`component > global > default`). Because `global` short-circuits the `default` branch, setting *any* key in `global.securityContext` drops the chart's defaults entirely, even keys the user never touched.

For example, the GPU host-access container loses `privileged: true` and 
`Watcher` and `supervisor` lose their hardened defaults

Suggested fix: Switch the helper to merge semantics so that chart defaults are preserved.

Also worth adding this scenario to the tests as it does not cover it.

Moran-k (commented):
Overall, this change looks good, one important comment below.

charts/komodor-agent/templates/_security.tpl
Moran-k: nice, the defaults are preserved now 👍

One thing, though, `mergeOverwrite` only deep-copies the default, so nested overrides on one component leak into others through `Values`.

example:
```bash
  helm template t charts/komodor-agent --set apiKey=t --set clusterName=t \
    --set 'global.securityContext.capabilities.drop[0]=ALL' \
    --set 'components.komodorAgent.watcher.securityContext.capabilities.add[0]=NET_BIND_SERVICE' \
    --show-only templates/deployment.yaml
```

The supervisor ends up with capabilities.`add: [NET_BIND_SERVICE]` too, even though only the watcher had that override. The watcher's merge mutates.Values.global.securityContext.capabilities are in place, and the supervisor picks it up on the next invocation.

The fix is just wrapping the other sources in deepCopy in `templates/_security.tpl` (lines 6 and 19), e.g.:
  mergeOverwrite (deepCopy $defaultSecurityContext) (deepCopy $globalSecurityContext) (deepCopy $securityContext)

would also be good to add a test asserting a watcher-only nested override doesn't show up on the supervisor.

davidcollom: Thanks for the catch, all really good ideas/catches ❤️  - This should be resolved in [f25925d](https://github.com/komodorio/helm-charts/pull/581/commits/f25925d41822b43bd3475bc64e975cf65d9d2993) 🙏

charts/komodor-agent/examples/hardened-values.yaml
Moran-k, line 1: Sorry for all the comments, I think this pr is very good!

Two things after reviwing the otel hardening:

**1. the `otelInit.volumes` override is at the wrong path**

  ```yaml
  opentelemetry:
    otelInit:
      volumes:
        varlogpods: {}
        varlibdockercontainers: {}
```
nothing reads `otelInit.volumes`. the templates read from `opentelemetry.volumes.varlogpods` / `opentelemetry.volumes.varlibdockercontainers` (see `templates/opentelemetry/_volumes.tpl` and `_containers.tpl`). 
so this block is a no-op,  `helm template -f hardened-values.yaml` still mounts
  `/var/log/pods` and `/var/lib/docker/containers` into otel-collector.

** 2. The new otel-collector securityContext conflicts with what it actually does
  The old comment was (which i think is right):
```
  # Must run as root to read host log files
  runAsUser: 0
  runAsGroup: 0
  runAsNonRoot: false
  capabilities.add: [DAC_OVERRIDE]
```

Now it's `runAsUser: 1001` / `runAsNonRoot: true` / `DAC_OVERRIDE` removed, but the host log volumes are still mounted (because of item #1), and those files are usually owned by root. 
The otel pod will start, but once it tries to read it, it will get EACCES at runtime.

if the goal is to "drop host log collection in the hardened example", put the volume override at `opentelemetry.volumes.varlogpods: {}` / or set `capabilities.telemetry.deployOtelCollector=false`) and verify the rendered DaemonSet doesn't have the volumeMounts.

if the goal is "keep host log collection", otel-collector needs to stay with `runAsUser: 0` + `DAC_OVERRIDE`.

also `test_hardened_values_linux_container_security_contexts` only checks securityContext fields, so it doesn't catch this, would be good to also assert that var/log/pods/var/lib/dockercontainers aren't mounted into a non-root container. (but this is nice to have)

Comment (davidcollom) [reply charts/komodor-agent/examples/hardened-values.yaml:1]: Don't be sorry! 😄 - I appreciate the feedback! Its all valid and great to see the attention to detail! ❤

File: .buildkite/tests/values_components_test.py [hunk 1]
```diff
@@ -255,6 +259,260 @@ def test_override_pod_security_context(component_name, resource_kind, resource_n
...
+    rendered_value = security_context
+    for key in field_path:
+        rendered_value = rendered_value[key]
+
+    if isinstance(rendered_value, list):
+        assert expected_value in rendered_value, \
+            f"Expected {expected_value} in {field_path}, got {security_context}"
+    else:
+        assert rendered_value == expected_value, \
+            f"Expected {field_path}={expected_value}, got {security_context}"
```
Moran-k: This and some other tests are indexed by `containers.3....` (positional). 
If someone reorders or adds a container in `daemonset.yaml`, this test breaks silently against the wrong container. 

Worth either iterating with find over containers by name == "otel-collector", or asserting containers[3].name == "otel-collector" before checking its securityContext

i guess we have this in many tests, so maybe we should fix all those in a separate PR

Moran-k (approved):
LGTM, minor comment but not a blocker

.buildkite/tests/values_components_test.py
davidcollom: Yea, I was thinking that too 🙈  it's not great if the ordering changes.... I'll update this test -happy to raise a Clickup for further review later or if I have time next week look at updating these more widely

PR #584, opened 2026-05-22
CU-86c9y1utw - ci: validate nginx version sync and auto-build on Dockerfile change

Moran-k:
## Summary

- Makes `scripts/nginx/Dockerfile` the single source of truth for the nginx version used by `komodor-agent`'s `komodorKubectlProxy`. `build-and-push.sh` now extracts the version from the Dockerfile and calls `komo ci docker-login` so it can run unattended in CI.
- Adds `scripts/nginx/validate-version.sh` — fails CI if the `FROM nginx:<tag>` in the Dockerfile drifts from `.components.komodorKubectlProxy.image.tag` in `charts/komodor-agent/values.yaml`.
- Wires two Buildkite steps into `.buildkite/pipeline.yml`:
  - **Validate nginx version sync** — runs on every build; cheap drift guard in both directions.
  - **Build & push nginx image** — `depends_on` the validator, only fires when the PR diff touches `scripts/nginx/Dockerfile`.

Future nginx bumps are one-line edits to `scripts/nginx/Dockerfile` + the matching `values.yaml` tag; CI publishes the image and fails if either side is forgotten.

## Note on publish scope

The `Build & push nginx image` step is **intentionally not gated on `build.branch == "master"`** — by design, a new nginx version is built and pushed **both** on the PR branch (every commit that touches `scripts/nginx/Dockerfile`) and on master after merge. This lets reviewers test the freshly published image during the PR cycle. The tag is immutable per-version, so re-pushing the same `1.30.x-alpine…` content is idempotent.

## Test plan

- [x] Buildkite: `Validate nginx version sync` step passes on this PR (Dockerfile and values.yaml both at `1.30.1-alpine3.23-slim`).
- [x] Buildkite: `Build & push nginx image` step triggers (Dockerfile is in the diff) and publishes `public.ecr.aws/komodor-public/nginx:1.30.1-alpine3.23-slim` + `komodorio/nginx:1.30.1-alpine3.23-slim`.
- [x] Manually verify negative case: push a follow-up commit that bumps the Dockerfile only — validator step should fail with a diagnostic naming both files.
- [x] Manually verify negative case (other direction): bump only `values.yaml` — validator should fail too.
- [x] Existing helm-template / readme / test suites still pass.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,32 @@ steps:
     if: build.message !~ /feat\(OSS.+\):/i
     retry: *retry
 
+  - label: ":eyeglasses: Validate nginx version sync (Dockerfile ↔ values.yaml)"
+    key: validate-nginx-version
+    command: ./scripts/nginx/validate-version.sh
+    agents:
+      builder: "builder-dind-agent"
+    if: build.message !~ /feat\(OSS.+\):/i
+    retry: *retry
+
+  - label: ":docker: Build & push nginx image (on Dockerfile change)"
+    key: build-push-nginx
+    depends_on: validate-nginx-version
+    commands:
+      - |
+        BASE_BRANCH="${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-master}"
+        git fetch -q origin "${BASE_BRANCH}"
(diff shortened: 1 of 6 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86c9y1utw](https://app.clickup.com/t/86c9y1utw)

MorLiberty (approved):

PR #589, opened 2026-05-25
CU-86c9nh6hq - add skills and integrations crds

shay-cohen:
# Komodor Agent — Klaudia Integration & Skill Sync (Helm)

## What this PR adds

Two new Custom Resource Definitions and the wiring needed for the agent to manage them:

| Kind | Group | Plural | Purpose |
|---|---|---|---|
| `MCPIntegration` | `app.komodor.com/v1alpha1` | `mcpintegrations` (`mcpi`) | Declaratively manage Klaudia MCP integrations from a cluster. |
| `Skill` | `app.komodor.com/v1alpha1` | `skills` (`ks`) | Declaratively manage Klaudia skills referenced by integrations. |

The agent (k8s-watcher) reconciles these CRs against the Komodor public-API. The chart-side change is everything needed to make that controller deployable: CRDs, config, RBAC, secret, validation.

---

### New

- `charts/komodor-agent/crds/mcpintegration.yaml` — `MCPIntegration` CRD.
- `charts/komodor-agent/crds/skill.yaml` — `Skill` CRD.
- `charts/komodor-agent/templates/secret-public-api-key.yaml` — Secret rendered from `publicApiKey` value (skipped if `publicApiKeySecret` is provided).

### Modified

- `charts/komodor-agent/values.yaml` — top-level `publicApiKey*` keys + `capabilities.klaudiaIntegrationSync` block.
- `charts/komodor-agent/templates/configmap.yaml` — renders the four new sync settings into the agent ConfigMap.
- `charts/komodor-agent/templates/clusterrole-k8s-watcher.yaml` — RBAC for the new resources + `secrets:get`.
- `charts/komodor-agent/templates/_helpers.tpl` — `komodorAgent.publicApiKey.secret.name` and `komodorAgent.publicApiKeySecretRef`.
- `charts/komodor-agent/templates/validations.yaml` — fail-fast if the capability is enabled without a public API key.
- `charts/komodor-agent/templates/watcher/_containers.tpl` — injects `KOMOKW_PUBLIC_API_KEY` into the watcher container when the capability is enabled.

---

## CRDs

Both CRDs live in `charts/komodor-agent/crds/`. Helm guarantees `crds/` are applied **before** any rendered manifest, so the controller can register informers on first start. CRDs in `crds/` are also intentionally **not removed** on `helm uninstall` (they need `kubectl delete crd ...` manually) — this is the standard chart contract and protects users from losing CR data on accidental uninstall.

### `MCPIntegration`

- `subresources.status: {}` → controller writes status without bumping `metadata.generation`.
- Printer columns: `Phase`, `IntegrationID`, `Tools` (count), `Age`.
- Required spec fields: `name`, `mcp_server`, `connectivity`.
- CEL validations enforce:
  - exactly one of `skill_id` / `skill_ref`,
  - per-`auth.method` block presence (`static_token` / `oauth2_client_credentials` / `token_exchange`),
  - `value` XOR `value_ref` for token references,
  - `agent_tunnel` only when `connectivity.mode == agent-tunnel`,
  - `secret_refs[].path` restricted to `mcp_server.headers.*` (extension point — backend currently uses this only for header injection).
- `status` is mostly opaque (`x-kubernetes-preserve-unknown-fields: true`) plus an explicit `toolCount` integer used by the printer column.

### `Skill`

- Required: `name`, `description`, `instructions` (with `maxLength` caps: 200 / 2000 / 50000).
- `use_cases` enum: `[chat, rca]` (default both).
- Printer columns: `Phase`, `SkillID`, `Age`.

Both CRDs are namespaced. `MCPIntegration.spec.skill_ref` resolves a sibling `Skill` in the same namespace.

---

## values.yaml — new keys

```yaml
# Top-level (NOT under capabilities — reusable by any feature that needs it later)
publicApiKey:                 # (guid) Komodor public API key
publicApiKeySecret:           # (string) existing Secret name (alternative to publicApiKey)
publicApiKeySecretKey:        # (string) key inside publicApiKeySecret (default: publicApiKey)

capabilities:
  klaudiaIntegrationSync:
    enabled: false             # (bool)   feature flag
    serverHost: ""             # (string) override Klaudia API base URL (empty = default)
    resyncInterval: "2m"       # (string) informer resync — re-enqueue safety net
    httpTimeout: "90s"         # (string) HTTP client timeout for all Klaudia calls
    serverFetchInterval: "20m" # (string) refresh server-side status; "0s" disables gating
```

### Why a separate "public" API key?

The agent's existing `apiKey` is the agent identity used for telemetry. Klaudia's public-API endpoints accept a customer-scoped **public** API key. Keeping them as separate values:

- avoids broadening the trust scope of the existing agent key,
- lets users rotate them independently,
- makes the public key reusable by future capabilities (hence top-level, not nested under `klaudiaIntegrationSync`).

### Why `serverFetchInterval` is separate from `resyncInterval`

| Knob | What it controls |
|---|---|
| `resyncInterval` (2m) | Informer re-list against the **K8s API server**. Re-enqueues every CR — no Klaudia traffic by itself. Cheap. |
| `serverFetchInterval` (20m) | When a CR has nothing local to do, controls how often we still hit **Klaudia** (`GET integration` / `GET skill`) to detect upstream drift / deletions. |

This split keeps Klaudia call volume bounded (≤ 1 per CR per 20m by default) while still letting the local controller wake up frequently enough to recover from missed events.

---

### `templates/secret-public-api-key.yaml` (new)

```yaml
{{- if and ((.Values.capabilities).klaudiaIntegrationSync).enabled (not .Values.publicApiKeySecret) }}
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "komodorAgent.publicApiKey.secret.name" . }}
type: Opaque
data:
  {{ default "publicApiKey" .Values.publicApiKeySecretKey }}: {{ .Values.publicApiKey | required "..." | b64enc }}
{{- end }}
```

- Skipped entirely if the capability is disabled.
- Skipped if the user provides their own `publicApiKeySecret` (BYO secret).
- Uses `( ).` index-style chaining (`((.Values.capabilities).klaudiaIntegrationSync).enabled`) so it doesn't blow up when the capability block is absent in stripped-down values overrides.

### `templates/_helpers.tpl` (added)

- `komodorAgent.publicApiKey.secret.name` → `<release>-komodorAgent-public-api-key-secret`.
- `komodorAgent.publicApiKeySecretRef` → returns the full `valueFrom.secretKeyRef` block, picking BYO vs chart-managed secret transparently. Used by `_containers.tpl`.

### `templates/configmap.yaml`

Adds a `klaudiaIntegrationSync` block to the agent config consumed by viper:

```yaml
klaudiaIntegrationSync:
  enabled: {{ ... }}
  {{- if .Values.capabilities.klaudiaIntegrationSync.serverHost }}
  serverHost: {{ ... | quote }}
  {{- end }}
  resyncInterval:        {{ ... | quote }}
  httpTimeout:           {{ ... | quote }}
  serverFetchInterval:   {{ ... | quote }}
```

`serverHost` is conditionally rendered so the agent default applies when unset.

### `templates/clusterrole-k8s-watcher.yaml`

Appended only when the capability is enabled:

```yaml
- apiGroups: ["app.komodor.com"]
  resources: ["mcpintegrations", "mcpintegrations/status", "skills", "skills/status"]
  verbs: ["get", "list", "watch", "patch", "update"]
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get"]
```

- `patch` is needed for finalizers (added/removed via `Update` on the unstructured object) and for status writes.
- `secrets:get` is required because integration `auth.*.value_ref` and `secret_refs[]` resolve Kubernetes Secrets at reconcile time. Watch is **not** required (resolved on demand).

### `templates/validations.yaml`

Fails the install/upgrade if `klaudiaIntegrationSync.enabled=true` but no public API key is provided, before any resources hit the cluster:

```yaml
{{- if ((.Values.capabilities).klaudiaIntegrationSync).enabled }}
  {{- if not .Values.publicApiKeySecret }}
    {{- if not .Values.publicApiKey }}
      {{- fail "publicApiKey (or publicApiKeySecret) is required when capabilities.klaudiaIntegrationSync.enabled is true" }}
    {{- end }}
  {{- end }}
{{- end }}
```

### `templates/watcher/_containers.tpl`

```yaml
{{- if ((.Values.capabilities).klaudiaIntegrationSync).enabled }}
- name: KOMOKW_PUBLIC_API_KEY
  {{ include "komodorAgent.publicApiKeySecretRef" . | nindent 4 }}
{{- end }}
```

Env var name is feature-neutral (`KOMOKW_PUBLIC_API_KEY`), not `KOMOKW_KLAUDIA_*`, because the value isn't Klaudia-specific.

---

## Backwards compatibility

- `enabled: false` is the default — existing installs see **no behavior change**, no new resources, no extra RBAC, no Secret.
- No existing values were renamed or removed.
- CRDs are additive. They live in `crds/` and won't be touched by `helm upgrade` after the first install (per Helm's CRD policy).

---

## How to verify locally

```bash
# Render with capability disabled (default) — should not contain any of the new resources
helm template ./charts/komodor-agent | grep -E "MCPIntegration|Skill |public-api-key" || echo "OK (none rendered)"

# Render with capability enabled
helm template ./charts/komodor-agent \
  --set apiKey=fake \
  --set publicApiKey=fake \
  --set capabilities.klaudiaIntegrationSync.enabled=true \
  --set clusterName=lab \
  | grep -E "klaudiaIntegrationSync|KOMOKW_PUBLIC_API_KEY|public-api-key-secret|mcpintegrations"

# Validation should fail when capability is on but no key is provided
helm template ./charts/komodor-agent \
  --set apiKey=fake \
  --set capabilities.klaudiaIntegrationSync.enabled=true \
  --set clusterName=lab
# Expected: Error: publicApiKey (or publicApiKeySecret) is required ...

# Lint
helm lint ./charts/komodor-agent --set apiKey=fake --set clusterName=lab
helm lint ./charts/komodor-agent --set apiKey=fake --set publicApiKey=fake \
  --set capabilities.klaudiaIntegrationSync.enabled=true --set clusterName=lab

# Apply CRDs against a kind cluster and try a sample CR
kubectl apply -f charts/komodor-agent/crds/
kubectl apply -f - <<EOF
apiVersion: app.komodor.com/v1alpha1
kind: Skill
metadata: { name: demo, namespace: default }
spec:
  name: "Demo"
  description: "Demo"
  instructions: "Use the demo MCP."
EOF
```

---

## Notes for reviewers

- The CRD field naming is **snake_case** (`skill_id`, `mcp_server`, `value_ref`) on purpose — these match the Klaudia backend Pydantic models 1:1, which keeps the agent payload builder a near-direct copy. We accept the snake_case style on CRD specs for that reason.
- `Skill.spec.use_cases` is restricted to `[chat, rca]` to mirror what the backend currently accepts. New use cases require a CRD change.
- `MCPIntegration.spec.secret_refs[].path` only allows `mcp_server.headers.*` today; the schema is structured to widen this list later without breaking changes (e.g. add `auth.headers.*`).
- All conditional logic uses defensive `((.Values.capabilities).klaudiaIntegrationSync).enabled` indexing so this chart is still renderable with stripped/minimal values files, e.g. in CI matrix tests that only set `apiKey`.

---

## Out of scope

- Migration from any existing UI-managed integrations into `MCPIntegration` CRs (manual; the agent will adopt by `name + managed_by=cluster:<clusterName>` so it's safe to point a CR at an existing row only when the row already has a matching `managed_by`).
- A dedicated CRD for chart-managed clusters list (`spec.clusters` is a free-form string array today; default `["*"]`).


charts/komodor-agent/values.yaml
Moran-k, line 1: you need to generate the readme `make -C charts/komodor-agent generate-readme`

charts/komodor-agent/crds/skill.yaml
```diff
@@ -0,0 +1,65 @@
...
+  name: skills.app.komodor.com
+spec:
+  group: app.komodor.com
+  scope: Namespaced
+  names:
+    plural: skills
+    singular: skill
+    kind: Skill
+    shortNames:
+      - ks
```
Moran-k: I think the name `skill` is too generic, maybe we should name it something like klaudiaSkills
Now it is still easy to change, once it is out, we are stuck with it.

something like:
```yaml
spec:
    group: app.komodor.com
    names:
      plural: klaudiaskills
      singular: klaudiaskill
      kind: KlaudiaSkill
      shortNames:
        - klskill
```

same for the `mcpintegrations`

charts/komodor-agent/templates/clusterrole-k8s-watcher.yaml
```diff
@@ -530,4 +530,12 @@ rules:
     - get
     - watch
     - list
+{{- if .Values.capabilities.klaudiaIntegrationSync.enabled }}
+  - apiGroups: ["app.komodor.com"]
+    resources: ["mcpintegrations", "mcpintegrations/status", "skills", "skills/status"]
+    verbs: ["get", "list", "watch", "patch", "update"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get"]
```
Moran-k: This creates a cluster-wide permission to get secrets, 

Is the intent that `MCPIntegration` CRs work in any namespace, or only the agent's namespace? If the latter, we should move the rbac to `Role + RoleBinding`.

shay-cohen:
@claude

Moran-k (approved):

PR #591, opened 2026-05-27
feat: otel-collector filelog IOPS safety controls

MorLiberty:
## Summary
Chart-level defense-in-depth for the otel-collector log-tailing IOPS issue (companion to the agents-service guardrail). Adds a documented namespace-exclude valve and a `start_at` safety note to the static filelog config.

## Context
The otel-collector `filelog` receiver tails container logs; its read rate scales with the log-write rate of the in-scope pods. With a broadened scope, on storage that doesn't benefit from the Linux page cache (VMware datastores, memory-pressured nodes) this becomes sustained read IOPS — the spike reported by Uline/Rockwell after upgrading to 2.17.0.

## What changed
`charts/komodor-agent/`
- **`values.yaml`** — new `components.komodorDaemon.opentelemetry.logCollection.excludeNamespaces` (default `[]`), with docs explaining the IOPS cost of broad log collection and that the live scope is backend-driven when `otelInit.enabled=true`.
- **`templates/opentelemetry/configmap.yaml`** — renders the excluded namespaces as filelog `exclude` globs; adds a comment that `start_at` must stay `end`.

## Notes
- Behavior-neutral by default (`excludeNamespaces: []`). Applies to the static fallback config used when `otelInit.enabled=false`; for remote-config agents the scope is enforced server-side (see mono PR #28601).

## Test plan
- [x] `helm lint` passes
- [x] `helm template` renders the exclude globs correctly for given namespaces


MorLiberty:
cu-86c9zyqaj

PR #601, opened 2026-06-07
CU-86ca5evq0 - feat(komodor-agent): add extraVolumes support to watcher container

shay-cohen:
Allow projected ServiceAccount tokens (and any custom volumes) to be
mounted into the watcher container via components.komodorAgent.watcher.extraVolumes.
Removes vendor-specific example from docs; adds CI coverage.


shay-cohen:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86ca5evq0](https://app.clickup.com/t/86ca5evq0)

Moran-k (approved):

PR #610, opened 2026-06-17
CU-86caakhhv - fix(komodor-agent): ship pod logs under hardened values

nirbenator:
## What

The hardened example values (`charts/komodor-agent/examples/hardened-values.yaml`) silently disabled OpenTelemetry **pod log collection**. Two things combined to break it:

1. The otel-collector container was pinned to non-root (`runAsUser/runAsGroup/runAsNonRoot: 1001`).
2. The host-log volumes were nulled (`varlogpods: null`, `varlibdockercontainers: null`), so the `filelog` receiver had nothing to read.

Net effect: with the hardened profile, no pod logs reached the Komodor backend / Datadog.

## Change

- Remove `runAsUser` / `runAsGroup` / `runAsNonRoot` from the otel-collector `securityContext`. The container then falls back to the chart-default context so it can read the host log files.
- Un-null `varlogpods` so `/var/log/pods` is mounted again.
- Keep `varlibdockercontainers` disabled (the legacy docker path is absent on containerd nodes).
- The collector still runs with `drop: ALL` capabilities (no `DAC_OVERRIDE`), `readOnlyRootFilesystem: true`, and `allowPrivilegeEscalation: false`.

```diff
     opentelemetry:
-      # Disable host log collection in the hardened profile...
+      # Keep pod log collection working under the hardened profile by mounting /var/log/pods...
       volumes:
-        varlogpods: null
         varlibdockercontainers: null

       securityContext:
-        runAsUser: 1001
-        runAsGroup: 1001
-        runAsNonRoot: true
         allowPrivilegeEscalation: false
         readOnlyRootFilesystem: true
         capabilities:
           drop: [ALL]
           add: []
```

## Test results

Deployed the hardened profile to a local **kind** cluster and verified logs end-to-end in Datadog.

```bash
kind create cluster --name test-logs-nir
helm install komodor-agent ./charts/komodor-agent \
  -f ./charts/komodor-agent/examples/hardened-values.yaml \
  --set apiKey=*** --set clusterName=test-logs-nir \
  --namespace komodor --create-namespace
```

**Iteration 0** — securityContext keys removed only (volumes still nulled):
- otel-collector came up healthy, but with **no `/var/log/pods` mount** → filelog receiver had nothing to read.
- Datadog `@clusterName:test-logs-nir` (otel-forwarded): **0 logs**. ❌

**Iteration 1** — also restored the `varlogpods` mount:
- Collector logs confirm it now reads host pod logs, no export errors:
  ```
  msg":"Started watching file" ... "path":"/var/log/pods/komodor_komodor-agent-.../0.log"
  ```
- Live container securityContext / mounts:
  ```
  securityContext: allowPrivilegeEscalation:false, readOnlyRootFilesystem:true,
                   capabilities:{drop:[ALL]}, runAsUser:0, runAsGroup:0, runAsNonRoot:false
  volumeMounts: opentelemetry-shared-config -> /etc/otel
                opentelemetry-varlogpods     -> /var/log/pods
  ```
- **Datadog confirmed** — agent pod logs arriving via the collector (`host: agent-otel-collector-*` = Komodor OTLP ingestion), tagged `@clusterName:test-logs-nir`:
  ```
  2026-06-17T11:30:12Z  info  admission-controller  successfully initialized komodor admission-controller [controller]
  2026-06-17T11:30:11Z  info  admission-controller  using cluster name from input
  ```
  ✅ Logs ship under hardened values.

Test cluster torn down after verification (`kind delete cluster --name test-logs-nir`).

## Note for reviewers

Because Helm deep-merges this overlay onto the chart default `values.yaml`, removing the three container keys makes the collector fall back to the default **root** context (`runAsUser: 0`). It still drops all capabilities and uses a read-only root FS. If we'd rather keep the collector non-root in the hardened example, the alternative is to keep `runAsUser: 1001` and add `DAC_OVERRIDE` back so a non-root collector can read root-owned host logs — happy to switch if preferred.

### DIFF
--- a/.buildkite/tests/values_components_test.py
+++ b/.buildkite/tests/values_components_test.py
@@ -491,7 +491,8 @@ def test_default_otel_collector_security_context_supports_host_log_collection(fi
     ("DaemonSet", "-daemon", "spec.template.spec.containers.0.securityContext"),
     ("DaemonSet", "-daemon", "spec.template.spec.containers.1.securityContext"),
     ("DaemonSet", "-daemon", "spec.template.spec.containers.2.securityContext"),
-    ("DaemonSet", "-daemon", "spec.template.spec.containers.3.securityContext"),
+    # containers.3 (otel-collector) is intentionally root under the hardened profile to
+    # support host log collection — covered by test_hardened_otel_collector_* below.
     ("DaemonSet", "-daemon", "spec.template.spec.containers.4.securityContext"),
 ])
 def test_hardened_values_linux_container_security_contexts(resource_kind, resource_name_suffix, container_path):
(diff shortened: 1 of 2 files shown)


nirbenator:
### Follow-up: re-added `DAC_OVERRIDE` to the hardened otel-collector

Pushed `f6c47a05`. The otel-collector `securityContext` now does `drop: [ALL]` + `add: [DAC_OVERRIDE]` instead of `add: []`.

**Why:** the previous revision relied on the host log files under `/var/log/pods` being world-readable (true on the kind/containerd test node, which is why logs shipped). On nodes/runtimes where those root-owned files are *not* world-readable, a collector with no capabilities could fail to read them even as root. `DAC_OVERRIDE` lets it bypass the file permission bits, making hardened log collection portable across environments.

This is a strict superset of the read access already validated in the test results above (logs confirmed in Datadog under `@clusterName:test-logs-nir`), so the prior end-to-end result still holds — it only broadens where it works.

Moran-k (approved):

nirbenator:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86caakhhv](https://app.clickup.com/t/86caakhhv)

PR #613, opened 2026-06-18
CU-86caavcuf - ci: notify Slack on new komodor-agent release

OrBin:
## Motivation
We want visibility into komodor-agent chart releases in a dedicated Slack channel.

## Changelog
- Add a Buildkite step after "validate helm chart version updated" that posts the new chart/agent version to Slack via an incoming webhook.
- The webhook URL is injected from SSM at `/app/ci/slack/release-webhook` (via the `aws-ssm` plugin); version info comes from buildkite meta-data.
- Runs in parallel with the install steps and is gated to real master releases (same `if:` as the rest of the release flow).
- Scope: komodor-agent only — the release/validate/install flow is hardcoded to that chart.

## Testing coverage
- [ ] Manual testing — requires the SSM param + Slack webhook to be provisioned first

> **Note:** This step will fail until the `/app/ci/slack/release-webhook` SSM SecureString is created. Can add a graceful-skip if preferred.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -257,6 +257,22 @@ steps:
     if: build.message !~ /feat\(OSS.+\):/i && build.branch == "master"
   - wait
 
+  - label: ":mega: Notify Slack about new komodor-agent release"
+    commands:
+      - |
+        CHART_VERSION="$$(buildkite-agent meta-data get komodor-agent-version)"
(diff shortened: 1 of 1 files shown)


OrBin:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86caavcuf](https://app.clickup.com/t/86caavcuf)

Moran-k (approved):

PR #626, opened 2026-07-02
[skip ci] CU-86cahvv67 - test: swap configmap smoke test for deploy event

Moran-k:
## What
Replace `test_get_configmap_from_resources_api` in `basic_test.py` with `test_get_deploy_event_from_resources_api`.

## Why
Backend is removing ConfigMap event ingestion. Once shipped, the existing test will always get an empty response from `/resources/api/v1/configurations/config-maps/events/search` and fail.


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cahvv67](https://app.clickup.com/t/86cahvv67)

kobikadosh (approved):

PR #640, opened 2026-08-20
fix(komodor-agent): gate EKS MNG spot-avoidance label behind opt-in

Moran-k:
## Why
Default admission-controller `nodeAffinity` ships `eks.amazonaws.com/capacityType NotIn [SPOT]`. On EKS Auto Mode this label domain is restricted and `capacityType` is not a well-known key, so Karpenter rejects `NodeClaim` creation and node provisioning halts cluster-wide.

## What
- Remove the key from the default preferred `matchExpressions`.
- Add `components.admissionController.spotAvoidance.eksManagedNodeGroups` (default `false`). When true, the template appends the expression to every preferred entry in the effective `nodeAffinity`, preserving user overrides.

## Migration
MNG users who relied on the default to steer admission-controller pods off SPOT: set `components.admissionController.spotAvoidance.eksManagedNodeGroups: true`. Auto Mode leaves it `false` (default).

## Verified
- Default render: no `eks.amazonaws.com/capacityType`.
- Opt-in render: key appended.
- Edge cases render clean (`nodeAffinity` null, `{}`, required-only, custom preferred).
- `make validate-readme` passes.


Moran-k:
CU-86cb7y5jy

MorLiberty (approved):

orsher (approved):

PR #641, opened 2026-08-23
cu-86cb8ttp0 feat(pipeline): install komodor-agent on agentops-hosted-agents

MorLiberty:
Adds `agentops-hosted-agents` to the komodor-agent install step, alongside the existing `agentops-production` / `agentops-staging` / `agentops-demo` lines, plus the per-cluster override values that cluster requires.

## ⛔ Merge order: komodorio/build-tools#418 first

`update_agent.sh` starts with `komo ctx "${environment}"`, and komo has **no `agentops-hosted-agents` context** — #418 adds it. Because that script runs under `set -e` with every cluster in a single Buildkite step, merging this first does not fail one cluster: it takes down the komodor-agent install for komodor production, `agentops-production`, `agentops-staging`, `agentops-demo`, `eu-production` **and** the rc chart.

I could not determine whether merging #418 propagates `komo` to the Buildkite agent automatically or the agent image must be rebuilt — worth confirming before merging this.

## The override file is not optional

`production-values.yaml` is wrong for this cluster **in both directions**:

- `komodorDaemon` and `komodorMetrics` get tolerations for `komodor.io/sensitive` and `dedicated=ws-redis` — **neither taint exists here**
- `komodorAgent`, `admissionController` and `komodorKubectlProxy` get **no tolerations at all**

Every Karpenter NodePool on this cluster is tainted on purpose: `platform` carries `komodor.io/agentops-hosted-agents-core`, `agents-production` carries `agentops.komodor.com/env=production`. A workload tolerating nothing goes Pending rather than landing on whatever pool happens to be untainted — so with the base values alone, **every komodor-agent workload stays Pending**.

Rendered both ways rather than reasoned about:

| | daemon | metrics | agent / admission-controller / proxy |
|---|---|---|---|
| base only | `sensitive` + `ws-redis` (neither exists) | `sensitive` (doesn't exist) | **no tolerations** |
| base + override | tolerates all taints | core toleration | core toleration |

Two deliberate choices worth reviewing:

- **The daemon gets a blanket `operator: Exists`**, not the two taints spelled out. That's the chart's own default, and it stays correct when a pool is added later. An explicit pair would silently stop monitoring the next pool someone introduces — and a monitoring DaemonSet that misses a pool reports healthy while seeing nothing. This is the trap `production-values.yaml` already fell into.
- **The Deployments tolerate only the core taint.** Not tolerating the agents-pool taint is precisely what keeps them off capacity sized for customer agents, so no `nodeSelector` is needed.

## State of the cluster right now

I installed the chart by hand first (before knowing this pipeline was the mechanism), so komodor-agent **is already running** there — release `komodor-agent` rev 1, chart 2.17.8, 7 pods healthy, 2 daemon pods matching the 2 `platform` nodes.

That install is a stopgap, and this PR supersedes it: `update_agent.sh` reuses the live values only for its dry-run preview, while the real `helm upgrade` passes `production-values.yaml` + this override + `--set` flags. So the next master build overwrites my hand-rolled values with these — which is the point, and also **exactly why the override file has to land with the pipeline line**: without it that overwrite would push a healthy cluster into all-Pending.

Also dropped from what I installed: `imagePullSecret: docker-cfg-komodorio` comes in via `production-values.yaml` and names a secret that exists on **neither** cluster. It's harmless — all images come from `public.ecr.aws/komodor-public`, which is public, and `agentops-production` has run this way for 42 days — so I left it rather than special-casing it here.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -299,6 +299,7 @@ steps:
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-production agentops-production komodor-agent
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-staging agentops-staging komodor-agent
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-demo agentops-demo komodor-agent
+      - ./.buildkite/pipeline_scripts/update_agent.sh agentops-hosted-agents agentops-hosted-agents komodor-agent
      # Install komodor agent on EU-production cluster report to US-production cluster
       - ./.buildkite/pipeline_scripts/update_agent.sh eu-production eu-production "komodor-agent-us" "" "komodor-agent-report-to-us"
       - ./.buildkite/pipeline_scripts/update_agent.sh production production-rc-chart komodor-agent-rc "" "komodor-agent-rc" "rc"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -299,10 +299,25 @@ steps:
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-production agentops-production komodor-agent
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-staging agentops-staging komodor-agent
       - ./.buildkite/pipeline_scripts/update_agent.sh agentops-demo agentops-demo komodor-agent
-      - ./.buildkite/pipeline_scripts/update_agent.sh agentops-hosted-agents agentops-hosted-agents komodor-agent
      # Install komodor agent on EU-production cluster report to US-production cluster
       - ./.buildkite/pipeline_scripts/update_agent.sh eu-production eu-production "komodor-agent-us" "" "komodor-agent-report-to-us"
       - ./.buildkite/pipeline_scripts/update_agent.sh production production-rc-chart komodor-agent-rc "" "komodor-agent-rc" "rc"
+    agents:
+      builder: "builder-dind-agent"
+    plugins:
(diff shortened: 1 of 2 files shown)


MorLiberty:
cu-86cb62h31

Moran-k (approved):

MorLiberty:
## Restructured: the install is now its own Buildkite step

I originally appended the line to the step that installs **six** targets in one `set -e` shell — komodor production, `agentops-production`, `agentops-staging`, `agentops-demo`, eu-production-reporting-to-US, and the rc chart. Any single failure there aborts the rest, and this is the newest, least-proven target of the set.

What changed my mind was the precedent's timing:

| | |
|---|---|
| `komo ctx agentops-demo` added (build-tools#410) | **2026-07-18** |
| pipeline line for `agentops-demo` merged (#636) | **2026-08-16** — four weeks later |

That gap is presumably why nobody had to think about whether a build-tools merge reaches the Buildkite agent. This change has *minutes*, and I still cannot determine whether `komo` propagates automatically or the agent image must be rebuilt — the README installs it by clone + `pip install --editable`, and there is no build-tools reference in komodorio/infra.

So rather than rely on that, the coupling is gone: `agentops-hosted-agents` now installs in its own step with the same agent, plugin and `if` condition. Worst case is one red step and this cluster staying on its current release — instead of six clusters silently missing a komodor-agent version bump.

`komodorio/build-tools#418` is merged, so the ordering dependency itself is satisfied; this just removes the blast radius if the agent hasn't caught up yet.

Nothing else changed — same `update_agent.sh` invocation, same override values file.

PR #644, opened 2026-09-06
CU-86cbejekx - fix(komodor-agent): drop RBAC rules whose resource list is empty

Moran-k:
Five rules in `clusterrole-k8s-watcher.yaml` have a fixed `apiGroups`/`verbs` but a fully conditional `resources` list. Turn off every kind in one of those groups and the rule renders `resources: null`, which the API server rejects:

```
The ClusterRole "komodor-agent-k8s-watcher" is invalid:
* rules[13].resources: Required value: resource rules must supply at least one resource
* rules[14].resources: Required value: resource rules must supply at least one resource
```

This blocks any hardened install. Turning off `ingress`, `ingressClass` and `networkPolicy` is enough to hit it, since that empties both the `extensions` and the `networking.k8s.io` rule.

The fix wraps each of the five in `{{- if or <its keys> }}`, so the rule is dropped instead of emptied. Same shape as the `argoWorkflows` guard already in the file. Groups covered: core (15 keys), `batch`, `storage.k8s.io`, `extensions`, `networking.k8s.io`.

The core condition needs all 15 keys, so the file now aliases `{{- $allowed := .Values.allowedResources -}}` at the top and the five guards use it. Rule bodies are untouched, which keeps the diff at 11 lines.

## Backward compatibility

Rendered output is byte identical at chart defaults, with `allowReadAll=false`, with `examples/hardened-values.yaml`, and with every values file in the repo. No `values.yaml`, `Chart.yaml` or README change.

The admission controller self-signed cert and its `checksum/certs-webhook` annotation are regenerated on every `helm template` run and differ between two renders of the same tree, so those lines are excluded from the diff.

## Verification

`kubectl apply --dry-run=server` against a kind cluster, rendering with the three networking kinds off: master is rejected with 4 invalid rules, this branch is accepted. Same result with all 25 keys off.

New test class in `.buildkite/tests/validations_test.py`, 30 cases: each group disabled on its own, all groups at once, and one case per key that keeps that kind alone and asserts its rule still renders.

Mutation checked against the template:

- deleting any one of the five guards turns a case red
- flipping any `or` to `and` turns a case red
- dropping any single key from a guard's `or` list turns a case red, 25 of 25
- deleting the whole ClusterRole turns 30 cases red, so nothing passes vacuously
- reordering the verbs in a rule, which is semantically identical, stays green

The per-key cases render with `capabilities.helm.enabled=false`. That capability adds its own `"" -> secrets` read rule to the same ClusterRole, which otherwise satisfies the assertion even when the core guard has lost `secret`. That was the one mutation the first version of the test missed.

Suite goes from 23 to 53 cases, 7s to 13s. No cluster needed, `make validations_test` is the only target that runs without kind.

## Noticed, not fixed

- The `argoWorkflows` guard checks only `workflows` and `cronWorkflows`, but the body also emits `workflowtemplates` and `clusterworkflowtemplates`. Enabling only those two drops the whole rule.
- `allowedResources.clusterRole` is declared in `values.yaml` but never referenced in this template.

Part of komodorio/planning#230.

### DIFF
--- a/.buildkite/tests/validations_test.py
+++ b/.buildkite/tests/validations_test.py
@@ -1,3 +1,4 @@
+import yaml
 import pytest
 from config import API_KEY
 from helpers.utils import get_filename_as_cluster_name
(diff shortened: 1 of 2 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbejekx](https://app.clickup.com/t/86cbejekx)

shaicantor (approved):

PR #645, opened 2026-09-06
CU-86cbejqct - feat(komodor-agent): add a cost install profile

Moran-k:
Adds `--set profile=cost` to the komodor-agent chart: an install preset for a cost-only agent.

Stacked on #644 (HC-1). That has to land first — without it the cost profile cannot be installed at all, because turning off every kind in an apiGroup makes a ClusterRole rule render `resources: null` and the API server rejects it.

Closes two sub-issues of komodorio/planning#192, one commit each so any of them can be reverted alone.

| Commit | Issue | What |
| --- | --- | --- |
| `192a530` | planning#241 | add `capabilities.resourceInfo.enabled` |
| `e5da8cd` | planning#231 | the cost profile |
| `c011109` | planning#231 | reject an unrecognised profile instead of ignoring it |

One extra commit, `caa33b7`, is not part of any sub-issue: the nginx build step diffs against `origin/$BASE_BRANCH`, but Buildkite clones `--depth 1` (which implies `--single-branch`), so `git fetch origin <branch>` writes `FETCH_HEAD` and no `origin/<branch>` ref. The diff then fails with "unknown revision" on **any PR whose base is not the default branch** — every stacked PR. A master-based PR passes only because `origin/master` already exists from the clone. Fetching into the tracking ref explicitly fixes it. This is the first stacked PR to hit it; #644 is green on the same code.

## What the profile does

```bash
helm upgrade --install komodor-agent komodorio/komodor-agent \
  --set apiKey=<KEY> --set clusterName=<NAME> \
  --set profile=cost
```

|  | default | `profile=cost` |
| --- | --- | --- |
| grants on the agent ServiceAccount | 385 | **135** |
| `*/*` wildcard rules | 3 | **0** |
| distinct resource types granted | 72 | **25** |
| rendered objects | 31 | 26 |

The admission controller keeps its own `*/*` get rule on its own ServiceAccount — it exists for the cost flows and is out of scope for this epic.

Kept: metrics collection, the admission controller, HPA and KEDA right-sizing, and read access to the kinds the cost flows consume (node, namespace, pod, deployment, statefulSet, daemonSet, job, cronjob, metrics, rollout).

Turned off: actions, crActions, helm, RBAC management, temp tokens, event creation, pod logs, the node enricher, resource-info, agent telemetry and the OpenTelemetry collector, the WebSocket tunnel and kube-apiserver tunneling, the kubectl proxy, the Klaudia integration sync, the Windows daemonset, and the 39 resource kinds no cost flow reads.

Both renders were accepted by a real API server — `kubectl apply --dry-run=server` against kind, 31 objects at defaults and 26 under the profile, no errors.

## Why a value and not a `values-cost.yaml` file

`helm install -f values-cost.yaml` cannot reach a file inside a chart pulled from the helm repo. `.helmignore` has no `values-*` pattern so the file *is* packaged into the `.tgz`, but a bare relative path passed to `-f` resolves against the operator's own working directory, not the chart. The alternatives — a raw GitHub URL, or telling customers to copy-paste the file — carry hosting, chart-version-pinning and air-gap problems.

## Why this is not the profile design that was rejected before

A `profile:` mechanism was prototyped and rejected earlier. The reason still stands, but it applies to that implementation and not to the idea: that prototype **nulled** profile-controlled defaults so "unset" could be told apart from "explicitly false".

`installed-values.yaml` in the agent ConfigMap is a verbatim `toYaml .Values`, it is shipped to the backend at identify, and the backend type-asserts on it. `lookupInstalledValuesActions` returns `(false, false)` when `capabilities.actions` is missing or not a bool, and `AllowAgentActions` then falls through to `return true` — permitting all 23 mutating action types. A null `capabilities.admissionController.enabled` silently disables right-sizing the same way.

Nulling is not required. This helper **sets concrete values and never deletes a key**, so neither can happen. There is a test asserting `installed-values.yaml` contains no null anywhere under `capabilities` or `allowedResources`, in both profile states. Every other backend consumer was checked and behaves correctly on a present-`false`: kubectl-proxy, the RBAC reconciler's cluster filter, and komodor-cost's admission-controller check all distinguish missing from false, and the telegraf/admission remote-config readers only look at image tags.

The chart already does exactly this in production: `migrateHelmValues` rewrites `capabilities.helm` from a bool into a map with sprig `set`, and that rewrite reaches the backend payload.

## The Argo kinds stay on, and that is not an oversight

`resources-api` asks every cluster for `rollouts.argoproj.io`, `workflows.argoproj.io` and `cronworkflows.argoproj.io` whenever the matching CRD is installed — `crdBackedKinds` in `komodor_service/crd_support.go`, appended into `allKinds` in the reconciler.

An earlier version of this profile turned all of those off. That drops the `argoproj.io` RBAC, and with `allowReadAll: false` there is no catch-all left, so the agent's list returns **forbidden**. `parsers.go` swallows only the exact string `"no supported kubernetes resource found"`, which the agent produces solely when a GVR cannot be resolved — i.e. when the CRD is absent. A 403 is a different error, `GetParallelResultsFor` aborts the batch on it, and **komodor_service reconciliation stops entirely for that cluster**: nothing is ever marked deleted.

So `rollout`, `argoWorkflows.workflows` and `argoWorkflows.cronWorkflows` stay at their chart default. `common.ROLLOUT` is also in `RightsizableServiceKinds`, so turning it off contradicted the profile's own purpose. `workflowTemplates` and `clusterWorkflowTemplates` are gated separately in the ClusterRole and nothing requests them, so those two do come off. This costs 26 grants and is covered by a test that asserts the rendered ClusterRole — not just the config key.

Worth a reviewer's eye: the `allowedResources.rollout` guard also carries a `patch` verb, so keeping rollout on brings one mutating grant back. The agent will not use it under `capabilities.actions: false`. Splitting that verb out would change the rule for every existing install, so it is not in this PR.

## Two things worth a reviewer's attention

**1. A profile takes precedence over `--set`.** Values written during rendering win, so `--set profile=cost --set capabilities.kubectlProxy.enabled=true` leaves the proxy off. This is the one behaviour that differs from passing an equivalent values file, and it is why the profile only ever turns capabilities *off*: everything the cost flows need is left at its chart default, so an operator can still shrink the install further (`--set profile=cost --set capabilities.metrics=false` works), and `allowedResources.customReadAPIGroups` is left alone so CRD groups can be added alongside the profile.

**2. Every rendered template runs the helper first, and that is deliberate over-application.** Helm sorts templates deepest-path-first then reverse-alphabetically, so which file observes the profile first is an accident of directory naming. A probe template under `templates/zz-later/` reads `actions: "true"` and `memory: "8Gi"` in the **same render** where `configmap.yaml` already emits `allowReadingPodLogs: false`. Today the chart is correct only because `templates/watcher/service.yaml` happens to sort first and carries the include.

An earlier version of this PR put the include only in the 22 templates that read a profiled key, policed by a regex. That regex misses `{{- $caps := .Values.capabilities }}`, `index .Values...` and `toYaml .Values`, and any new subdirectory sorting after `watcher/` reopens the hole. So the include now goes in **all 41 rendered templates** and render order stops mattering at all.

Partials are excluded because helm never renders a file whose basename starts with `_` — which also means the existing `{{- include "migrateHelmValues" . -}}` at the top of `templates/watcher/_containers.tpl` is dead code (it only appears to work because `deployment.yaml` includes `configmap.yaml` for its checksum first). A test keeps the profile helper from growing the same dead wiring.

`values_profile_test.py` enforces both halves: every rendered template includes the helper before any other action, and no partial carries a top-level include. Dropping a probe template into `templates/zz-later/` without the include fails the suite; adding the include makes it pass and the probe then reads `actions: "false"`. Moving any include to the bottom of its file fails the suite.

## `capabilities.resourceInfo.enabled` (planning#241)

Written into the agent ConfigMap, not as a container env var. The ConfigMap is the agent's config file and the agent merges backend remote config at that same layer, so this stays a **default that Komodor can still override remotely**. An env var would beat remote config and would turn resource-info off under customers who had it enabled remotely.

## Watcher sizing is deliberately NOT in the profile (planning#243)

An earlier revision of this PR had the profile set the watcher memory limit to 2Gi. It is out again, and #243 is not closed by this PR.

The reasoning behind a lower limit still holds: the agent reads its own container memory limit and self-throttles at 20% remaining, so at 8Gi it does not start protecting itself until it is using 6.4Gi, by which point the node has OOM-killed it. 8Gi is sized for ~50 informers and a cost-profile watcher runs far fewer.

But a profile is the wrong place to enforce it. Values the profile writes take precedence over `--set` — correct for capability switches, which are the posture the profile exists to hold, and wrong for resource sizing, which belongs to the operator. The chart ships `utilities/memory-planning` whose entire output is a value to put here, so a customer must be able to set it.

The profile now writes no requests or limits at all, and a test asserts the watcher and supervisor resources are byte-identical with and without it. The recommendation moved into `README.md.gotmpl` next to the memory-planning section, as guidance:

```bash
--set profile=cost --set components.komodorAgent.watcher.resources.limits.memory=2Gi
```

## Known limitations of the profile

Both are documented in the README, not just here.

- Right-sizing keeps applying automatically through the admission controller, but `capabilities.actions: false` means **applying a recommendation by hand from the UI, or through self-healing, is refused** — `ActionTypeUpdateContainersResources` is one of the gated mutating action types. `capabilities.actions` is a single switch covering all 23, so there is no way to allow only that one.
- The agent still advertises live resource browsing for every kind: `enableAgentTaskExecution` is hardcoded `true` in the ConfigMap and the web gates `resourceList` / `liveDataView` on it alone. **Browsing a kind the profile no longer grants read on returns a permission error** rather than "not supported". Making that profile-controlled is a chart change beyond this PR.

Kinds the profile stops watching are not garbage-collected on the backend — the only soft-delete reconcilers are komodor_service, pod, node, namespace and workflow, and every source kind they read stays on. Existing rows for un-watched kinds persist and go stale rather than disappearing.

## Backward compatibility

With `profile` unset the rendered output is unchanged except for the two new `resourceInfo` lines, the new `profile` key in `installed-values.yaml`, and the resulting `checksum/config` annotation — so existing installs get one watcher restart on upgrade and nothing else. Verified by rendering before and after across chart defaults, `allowReadAll=false`, `examples/hardened-values.yaml`, `--set createRbac=false` and every values file in `.buildkite/pipeline_scripts/`, excluding the admission-controller cert lines which are regenerated on every render.

Per planning#231, MONO-2 (#233) or OPS-1 (#232) should land before this profile is applied to an **existing** cluster. New clusters have no rows to soft-delete; existing ones do.

## Known residue

Three read grants survive the cost profile because their ClusterRole rules are not keyed on `allowedResources` at all: `rbac.authorization.k8s.io/clusterroles`, `apps/replicasets` and `apiextensions.k8s.io/customresourcedefinitions` (the last one is deliberate — it is how the agent probes which CRDs exist). The agent stops *watching* all three, since collection follows the ConfigMap, but the permission remains. Fixing that means changing rules that render for everyone today, so it is not in this PR.

## Tests

`.buildkite/tests/values_profile_test.py`, 113 cases, `helm template` only, no cluster. 275 pass across every cluster-free suite in `.buildkite/tests`. `make -C charts/komodor-agent validate-readme` passes.

The suite was mutation-tested rather than trusted. Each of these makes it fail: dropping or inverting any `set` line, dropping a key from a resource list, putting `rollout` back in the off list, removing an include from any of the 22 templates, **moving an include to the bottom of its file**, adding a template that sorts after `watcher/` without the include, deleting either `kindIs "map"` guard, deleting the `migrateHelmValues` call, and deleting the `resourceInfo` block from the ConfigMap. Assertions against keys whose chart default already matches the profile are driven from a render that sets every profiled key `true` first, so none of them can pass vacuously.

One helm-4 note: `--set capabilities.helm=true` no longer replaces the declared map, so the legacy bool migration path can only be exercised on helm 3.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,7 +46,11 @@ steps:
       - |
         set -euo pipefail
         BASE_BRANCH="$${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-master}"
-        git fetch -q origin "$$BASE_BRANCH"
+        # Buildkite clones --depth 1, which implies --single-branch, so the only configured
+        # refspec is the default branch. `git fetch origin <branch>` then writes FETCH_HEAD
+        # but no origin/<branch> ref, and the diff below fails on any PR whose base is not
+        # the default branch (i.e. every stacked PR). Fetch into the tracking ref explicitly.
+        git fetch -q origin "+refs/heads/$$BASE_BRANCH:refs/remotes/origin/$$BASE_BRANCH"
         git fetch -q --unshallow 2>/dev/null || git fetch -q --deepen=100 || true
         # Capture diff outside the `if` so `set -e` reaches a `git diff` failure
         # (inside an `if` condition, set -e is disabled and the failure would be masked).
--- a/.buildkite/tests/values_profile_test.py
+++ b/.buildkite/tests/values_profile_test.py
@@ -1,0 +1,439 @@
+import os
+import re
+
+import pytest
+import yaml
+
+from config import API_KEY, CHART_PATH, RELEASE_NAME
+from helpers.helm_helper import helm_agent_template
+
+TEMPLATES_DIR = os.path.join(CHART_PATH, "templates")
+# the agent ConfigMap is named after the chart, everything else after the release
+CONFIG_MAP_NAME = "komodor-agent-config"
+FULLNAME = f"{RELEASE_NAME}-komodor-agent"
+AGENT_SERVICE_ACCOUNT = FULLNAME
+
+# Every value path "--set profile=cost" writes, as it appears in installed-values.yaml.
+# A path is listed here whether or not the profile's value differs from the chart default:
+# the profile pins it either way, and test_cost_profile_turns_off_every_profiled_key asserts
+# the rendered result rather than the delta.
+PROFILED_CAPABILITY_PATHS = [
+    "capabilities.actions",
+    "capabilities.crActions",
+    "capabilities.rbac",
+    "capabilities.rbacTempTokens",
+    "capabilities.helm.enabled",
+    "capabilities.events.create",
+    "capabilities.nodeEnricher",
+    "capabilities.logs.enabled",
+    "capabilities.resourceInfo.enabled",
+    "capabilities.telemetry.enabled",
+    "capabilities.telemetry.deployOtelCollector",
+    "capabilities.tunnel.enabled",
+    "capabilities.tunnel.kubeapiserver.enabled",
+    "capabilities.kubectlProxy.enabled",
+    "capabilities.klaudiaIntegrationSync.enabled",
+    "components.komodorDaemonWindows.enabled",
+]
+
+ALLOWED_RESOURCES_OFF = [
+    "allowReadAll",
+    "replicaSet", "horizontalPodAutoscaler", "podDisruptionBudget", "priorityClass",
+    "persistentVolume", "persistentVolumeClaim", "storageClass", "csiDriver", "csiNode",
+    "csiStorageCapacity", "volumeAttachment",
+    "limitRange", "resourceQuota", "event", "replicationController", "podTemplate",
+    "controllerRevision", "runtimeClass", "lease", "certificateSigningRequest",
+    "service", "endpoints", "endpointSlice", "ingress", "ingressClass", "networkPolicy",
+    "secret", "configMap",
+    "clusterRole", "clusterRoleBinding", "role", "roleBinding", "serviceAccount",
+    "customResourceDefinition", "admissionRegistrationResources", "authorizationResources",
+    "flowControlResources", "policyResources",
+]
+
+# Left at the chart default on purpose, so an operator can still shrink the install further.
+ALLOWED_RESOURCES_ON = [
+    "node", "metrics", "namespace", "pod",
+    "deployment", "statefulSet", "daemonSet", "job", "cronjob",
+    # resources-api requests rollouts.argoproj.io from every cluster whose CRD is installed
+    # (crdBackedKinds in komodor_service/crd_support.go). Without the read the agent answers
+    # forbidden, which the reconciler does not swallow, and reconciliation stops for the
+    # cluster. Rollout is also a rightsizable service kind.
+    "rollout",
+]
+
+# Same reason: workflows and cronworkflows are the other two crdBackedKinds. The remaining two
+# argo sub-keys are gated separately in the ClusterRole and nothing requests them.
+ARGO_WORKFLOWS_ON = ["workflows", "cronWorkflows"]
+ARGO_WORKFLOWS_OFF = ["workflowTemplates", "clusterWorkflowTemplates"]
+
+# What the cost flows need and the profile must not touch.
+COST_CAPABILITIES_ON = [
+    "capabilities.metrics",
+    "capabilities.admissionController.enabled",
+    "capabilities.cost.hpa",
+]
+
+# Reads that oblige a template to include "applyProfile" first. Matched against template text
+# with parentheses stripped, so "((.Values.capabilities).logs).enabled" matches too.
+PROFILED_READS = [
+    r"\.Values\.capabilities\.actions\b",
+    r"\.Values\.capabilities\.crActions\b",
+    r"\.Values\.capabilities\.helm\b",
+    r"\.Values\.capabilities\.rbac\b",
+    r"\.Values\.capabilities\.rbacTempTokens\b",
+    r"\.Values\.capabilities\.nodeEnricher\b",
+    r"\.Values\.capabilities\.logs\.enabled\b",
+    r"\.Values\.capabilities\.events\.create\b",
+    r"\.Values\.capabilities\.telemetry\.(enabled|deployOtelCollector)\b",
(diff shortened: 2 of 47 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbejqct](https://app.clickup.com/t/86cbejqct)

.buildkite/pipeline.yml
```diff
@@ -46,7 +46,11 @@ steps:
       - |
         set -euo pipefail
         BASE_BRANCH="$${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-master}"
-        git fetch -q origin "$$BASE_BRANCH"
+        # Buildkite clones --depth 1, which implies --single-branch, so the only configured
+        # refspec is the default branch. `git fetch origin <branch>` then writes FETCH_HEAD
+        # but no origin/<branch> ref, and the diff below fails on any PR whose base is not
+        # the default branch (i.e. every stacked PR). Fetch into the tracking ref explicitly.
+        git fetch -q origin "+refs/heads/$$BASE_BRANCH:refs/remotes/origin/$$BASE_BRANCH"
```
Moran-k, line 53: Bug-fix when stacking prs

MorLiberty (approved):

PR #647, opened 2026-09-07
CU-86cbewg52 - feat(komodor-agent): stop watching the workload kinds under profile=cost

Moran-k:
Stage 1 of komodorio/planning#256: the cost profile stops watching the workload kinds.

**Stacked on #645** (`mk/hc-2-cost-profile`), which introduces the profile. Review this one on its own two commits — the diff against master will look much larger.

| Commit | What |
| --- | --- |
| `23251a0d` | gate the `apps` ClusterRole rule on `allowedResources` |
| `5d551b43` | turn the eight kinds off in the profile |
| `3e24d4a7` | pin that `profile: cost` reaches the backend |

## What it does

`profile=cost` additionally sets `deployment`, `statefulSet`, `daemonSet`, `rollout`, `job`, `cronjob`, `argoWorkflows.workflows` and `argoWorkflows.cronWorkflows` to false.

| k8s-watcher ClusterRole | grants | resource types | wildcards |
| --- | --- | --- | --- |
| default install | 321 | 72 | 0 |
| `profile=cost` on #645 | 81 | 23 | 0 |
| **`profile=cost` here** | **40** | **11** | **0** |

What the agent still reads: pods, nodes, namespaces, replicasets, metrics, and the leases/CRD plumbing.

A default install renders byte-identically — verified across chart defaults, `allowReadAll=false`, `examples/hardened-values.yaml`, and all five values files under `.buildkite/pipeline_scripts/`.

## Why the first commit is needed

`daemonsets`, `deployments` and `statefulsets` were listed unconditionally in the `apps` rule. Turning the matching `allowedResources` key off stopped the informer but left the read grant, so without this the profile would shrink what the agent collects without shrinking what it can reach. Most of the 81 -> 40 reduction comes from here.

`replicasets` deliberately stays ungated. With every other kind in that rule off it is the only entry left, and it is what stops the rule rendering `resources: null`, which the API server rejects with *"resource rules must supply at least one resource"* while `helm template` still exits 0.

## Four of the six kinds are free

The backend has two ways to learn which service a pod belongs to: the agent reports it in the metric, or the backend looks it up in its own database, which the informers fill. It trusts the agent's value only for kinds on the `owner_ref_verified` allowlist (`metrics-collector/pkg/endpoints/metrics/handlers/podOwnerRef.go:138`).

`Deployment`, `StatefulSet`, `DaemonSet`, `Rollout` and `ArgoWorkflow` are all on that list. Their identity travels in the agent's own tag, so cost and right-sizing keep working with no informer at all.

Note this corrects a claim in #645, which kept `rollout` on because `common.ROLLOUT` is in `RightsizableServiceKinds`. That list is only a ClickHouse column filter (`metrics-api/pkg/cost/dal.go:342`) — it never performs a lookup.

## Two of them are a real trade

`Job`, `CronJob` and `ArgoCronWorkflow` are **not** on the allowlist. Without the informers their `komodor_service_kind` stays null, and null is filtered out at both `allocation/over_time/sql/ch.sql:52` and `shared_rows/sql/ch.sql:320-323`. So those pods:

- disappear from the service-level cost breakdown (no placeholder row — `dal.go:39` drops them on the null name as well)
- stop producing right-sizing recommendations, since `CronJob` is in `RightsizableServiceKinds` but can only reach that filter through the lookup
- keep their **node capacity** cost in the cluster total (`hourly_node_summary`, service-agnostic) but lose their **allocated** cost (`monthly_service_summary`, filtered at `20260507161656_...up.sql:238-239`)

Accepted deliberately: a cost cluster has no `komodor_service` data to look up either way.

**Live resource browsing for these kinds now returns a permission error.** All three ClusterRoles bind the same ServiceAccount, and the metrics roles keep `get` on the apps and batch kinds — so point reads still work, but any `list` 403s. `agent-task-manager/pkg/endpoints/resources_request/resources_request.go` builds a `TaskTypeResourceListV2` task from a caller-supplied kind, which is exactly that. #645 already documents this class of limitation; stage 1 extends it to the kinds people browse most. For the argoproj kinds the SA retains nothing at all, so both `get` and `list` fail.

**How much of this is actually new is unclear, and probably small.** The per-account enrichment gate `shouldEnrichTimestreamAccount` (`metrics-enricher/pkg/enricher/pod_ownership_fetcher.go:181-188`) **defaults to false** and returns false on error. For any account without it, these pods are already excluded from service cost and right-sizing today, with no chart change involved. The delta exists only for accounts where that flag is on.

One edge case worth naming: `mv_daily_cluster_summary` joins its allocation half with an `INNER JOIN` and carries a TODO about it (`:161-162`). A cost cluster running *only* Jobs and CronJobs would have no qualifying workloads and drop out of `daily_cluster_summary` entirely.

## Deploy-order dependency

The profile must not reach a backend older than komodorio/mono#30520. #645 kept the argo kinds on because Komodor asked every cluster for `workflows.argoproj.io` whenever the CRD was installed, and a forbidden answer aborted the whole `komodor_service` batch, leaving nothing marked deleted. #30520 makes resources-api skip every reconciler for a cost-profile cluster, which is what makes turning them off safe.

That guarantee now lives in mono rather than in this chart. The chart cannot test mono's half, but it does have a half of its own, and it was untested: **#30520 identifies a cost cluster by reading `installed-values.profile`.** If this chart ever stops emitting that value the filter silently stops matching, every reconciler resumes against cost clusters, and the 403 comes back — with a correctly-deployed backend and no deploy-order problem at all. `installed-values.yaml` is a verbatim dump of `.Values`, so it holds today by accident of `--set profile=cost` rather than by design. `3e24d4a7` pins it.

The remaining half is release ordering: don't publish a chart version carrying this until #30520 is live in both US and EU.

## Tests

183 pass. Six pre-existing tests encoded the old behaviour and were rewritten rather than deleted; `test_cost_profile_grants_read_on_the_argo_kinds` became `..._drops_the_argoproj_rules_entirely`, which is strictly stronger — it asserts the whole rule list is empty rather than three named resources.

Every assertion was mutation-checked. Each of these breaks a named test: removing any one of the six kinds from the off-list; removing either argo key; inverting a `set`; removing any of the three new ClusterRole gates; and misspelling a gate path.

That last one is the reason `apps` was added to `validations_test.py`'s group map. A misspelled `.Values` path renders falsy rather than erroring, so the cost render still looks correct — only the default-install render exposes it, and that map entry is the sole protection.

Two assertions were removed for being vacuous: `pods` and `nodes` could not fail under the cost profile, since `pods` survives via an ungated core rule and `nodes` via `metrics.k8s.io`. `validations_test.py` already covers both, scoped by apiGroup and verbs.

## Review

Three independent reviewers with no shared context: a chart reviewer, a backend blast-radius challenger, and a test auditor. No correctness defects were found. What they did find, and what changed as a result:

- two of my code comments were factually wrong about which ClusterRole grants what — all agent workloads share one ServiceAccount, and the metrics roles grant it `get` on the same apps/batch kinds, so what this actually removes cluster-wide is `list`/`watch`. Both comments fixed
- the two vacuous test entries above
- the cross-repo dependency on #30520 was implicit and is now written down

Two of their findings were rejected after checking the code. One reported phantom right-sizing rows and inflated savings, but traced two materialized views that later migrations dropped; the live path writes NULL where they wrote `''`, and NULL is filtered out rather than grouped. The other reported that cost-by-service was unaffected, but read the right-sizing rows query instead of the allocation query. Both were confirmed retracted by their authors.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,7 +46,11 @@ steps:
       - |
         set -euo pipefail
         BASE_BRANCH="$${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-master}"
-        git fetch -q origin "$$BASE_BRANCH"
+        # Buildkite clones --depth 1, which implies --single-branch, so the only configured
+        # refspec is the default branch. `git fetch origin <branch>` then writes FETCH_HEAD
+        # but no origin/<branch> ref, and the diff below fails on any PR whose base is not
+        # the default branch (i.e. every stacked PR). Fetch into the tracking ref explicitly.
+        git fetch -q origin "+refs/heads/$$BASE_BRANCH:refs/remotes/origin/$$BASE_BRANCH"
         git fetch -q --unshallow 2>/dev/null || git fetch -q --deepen=100 || true
         # Capture diff outside the `if` so `set -e` reaches a `git diff` failure
         # (inside an `if` condition, set -e is disabled and the failure would be masked).
--- a/.buildkite/tests/validations_test.py
+++ b/.buildkite/tests/validations_test.py
@@ -114,7 +114,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_missing_admission_controller_sa_name_when_not_creating(self):
         """Test that admission controller serviceAccount.name is required when create is false"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=true " \
+                   f"--set capabilities.admissionController.enabled=true " \
                    f"--set components.admissionController.serviceAccount.create=false"
         output, exit_code = helm_agent_template(settings=settings)
 
@@ -126,7 +126,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_admission_controller_sa_name_provided_when_not_creating(self):
         """Test that providing admission controller serviceAccount.name when create is false works"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=true " \
+                   f"--set capabilities.admissionController.enabled=true " \
                    f"--set components.admissionController.serviceAccount.create=false " \
                    f"--set components.admissionController.serviceAccount.name=my-ac-sa"
         output, exit_code = helm_agent_template(settings=settings)
@@ -137,7 +137,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_admission_controller_disabled_skips_validation(self):
         """Test that admission controller SA validation is skipped when disabled"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=false " \
+                   f"--set capabilities.admissionController.enabled=false " \
                    f"--set components.admissionController.serviceAccount.create=false"
         output, exit_code = helm_agent_template(settings=settings)
 
--- a/.buildkite/tests/validations_test.py
+++ b/.buildkite/tests/validations_test.py
@@ -230,6 +230,15 @@ class TestAllowedResourcesEmptyRuleValidation:
                 "serviceAccount": "serviceaccounts",
             },
         ),
+        "apps": (
+            ["apps"],
+            {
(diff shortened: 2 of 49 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbewg52](https://app.clickup.com/t/86cbewg52)

MorLiberty (approved):

PR #256, opened 2023-09-27
Helm values migration

Moran-k:
# Komodor Helm Migration

## Overview

This Go application is designed to automate the process of migrating Komodor k8s-watcher Helm chart values to a new structure. It fetches the existing Helm values for a given release, flattens the nested YAML, and then maps the old keys to new keys based on a predefined mapping.

## Prerequisites

- Go 1.16 or higher
- Helm 3.x
- Access to a Kubernetes cluster
- GNU Make

## How to Build

The application can be built for multiple platforms: Windows, Linux, and macOS (Darwin), targeting both `amd64` and `arm64` architectures.

1. Clone the repository:
2. Navigate to the project directory:

    ```bash
    cd scripts/helm-migration
    ```

3. Run the Makefile to build the application:

    ```bash
    make all VERSION=<VERSION>
    ```

This will produce the binaries in a `dist` directory. You can also build for specific platforms:

- For Windows:

    ```bash
    make windows VERSION=<VERSION>
    ```

- For Linux:

    ```bash
    make linux VERSION=<VERSION>
    ```

- For macOS (Darwin):

    ```bash
    make darwin VERSION=<VERSION>
    ```

To clean the `dist` directory, you can run:

```bash
make clean
```

## Usage

Run the application with the following command:

```bash
cd dist
./komodor-helm-migration_<OS>_<ARCH> -o [OUTPUT_FILE]
```

- `OUTPUT_FILE`: Optional. The name of the output file where the new Helm values will be saved. Default is `komodor_watcher_values.yaml`.

The application will:

1. Identify the namespace where the Helm release `k8s-watcher` is installed.
2. Fetch the existing Helm values for the release.
3. Map the old values to the new structure based on the predefined mapping.
4. Save the new values to the specified output file.
5. Print a Helm installation command using the new values file.

## ToDo:
Update the mapping

PR #648, opened 2026-09-07
CU-86cbexkdp - feat(komodor-agent): shrink the cost profile to zero informers, and make resourceInfo actually switchable

Moran-k:
Stages 2 and 3 of komodorio/planning#256: the cost profile stops watching nodes, pods and namespaces.

**Stacked on #647** (stage 1), which is stacked on #645. One commit per stage, so either can be reverted alone; the diff against master will look much larger.

## What it does

`profile=cost` additionally sets `allowedResources.node: false`, stopping the watcher's node informer. Watcher ClusterRole goes from 40 grants to 37 (distinct resource types stay at 11 — `nodes` still appears via the metrics-gated rule); a default install is unchanged.

The headline is not the grant count. On a large estate nodes are numerous and large objects, so not running that informer is the point — see komodorio/planning#243.

## No ClusterRole carve-out, contrary to the issue

komodorio/planning#238 and #256 both state that the k8s-watcher's core `nodes` grant is the only one, and that turning it off without a carve-out leaves `nodelabels` empty, resolves the provider to `KUBERNETES`, and silently zeroes a customer's enterprise discount.

That is not correct, and the reason is easy to miss: **three** ClusterRoles bind the agent's single ServiceAccount.

```
komodor-agent-k8s-watcher
komodor-agent-daemon-metrics          <- clusterrolebinding-daemon-metrics.yaml:13
komodor-agent-metrics-deployment
```

Both metrics roles name `komodorAgent.serviceAccountName` — the shared SA, not one of their own — and grant `namespaces`, `pods`, `nodes`, `nodes/stats`, `nodes/proxy` with `get`/`list`, carrying **no `allowedResources` gating at all**. They are gated only on `{{- if and .Values.createRbac .Values.capabilities.metrics }}`, and the cost profile keeps `capabilities.metrics` on by definition.

Effective permissions on the agent SA, computed per `(apiGroup, resource)` across every bound role:

| | before | after |
| --- | --- | --- |
| `core/nodes` | get, list, watch | **get, list** |
| `core/pods` | get, list, watch | unchanged |
| `core/nodes/stats`, `core/nodes/proxy` | get, list | unchanged |
| `metrics.k8s.io/nodes`, `metrics.k8s.io/pods` | get, list, watch | unchanged |

Only `watch` is removed. So `komodor-agent cmd/watcher/watcher.go:525` — the one-shot `Nodes().List(Limit: 1)` that populates `clusterregion` and `nodelabels` — still succeeds, and provider detection is unaffected.

Keying that calculation on the resource name alone conflates `core/pods` with `metrics.k8s.io/pods` and produces the opposite answer. The test below keys on both.

## The cost pipeline is protected by assertion, not by comment

The profile exists to shrink the agent without breaking cost collection, so that is now a test rather than a paragraph. `COST_RBAC_FLOOR` pins the ServiceAccount's effective permissions:

```
core/nodes            get, list
core/pods             get, list
core/namespaces       get, list
core/nodes/stats      get, list
core/nodes/proxy      get, list
metrics.k8s.io/pods   get, list
metrics.k8s.io/nodes  get, list
```

That covers telegraf reading kubelet and cAdvisor, per-node pricing (which resolves from the node object telegraf sends rather than from `ClusterSummary.Provider`), and the agent's own startup node list.

The admission controller is untouched by any of this — it runs on its own ServiceAccount whose ClusterRole carries no `allowedResources` gating, so right-sizing and its `*/*` get are unaffected by every stage of this work.

## Tests

191 pass. Two mutations, each caught:

| mutation | caught by |
| --- | --- |
| `node` not actually added to the off-list | `turns_off_the_unused_resource_kinds[node]` + `keeps_node_reads_despite_turning_the_node_kind_off` |
| **`nodes` removed from the two metrics ClusterRoles** | `keeps_the_rbac_the_cost_pipeline_needs[-nodes-*]` + the same node-reads test |

The second is the one that matters. It is the silent break this change depends on not happening — if someone later gates the metrics roles on `allowedResources`, the agent stops being able to read nodes and the discount zeroes with no error anywhere. Before this PR nothing would have caught that.

## Backend dependency, already satisfied

komodorio/mono#30521 is live in US and EU (`workspaces-api 1.1234.19532`, which contains its merge commit). It makes provider detection fall back to the agent's reported node labels when the state API returns no nodes — which is the state a cost cluster is in once its node informer stops. Not strictly required given the RBAC above, but it is what makes the no-nodes case correct rather than lucky.

---

# Stage 3: pods and namespaces (added to this PR)

`profile=cost` now also sets `allowedResources.pod: false` and `allowedResources.namespace: false`. That leaves **`metrics` as the only kind still on** — telegraf, which is the cost data itself — so the watcher runs zero resource informers on a cost cluster.

## No attribution loss — measured, not assumed

komodorio/planning#256 flagged stage 3 as "to be confirmed — do not assume". Confirmed on a live cluster, and the answer is better than the issue expected: **cost by service keeps working with zero informers.**

A kind cluster installed from scratch with `profile=cost`, which has therefore never run a pod, namespace or node informer, returns 9 named services with correct kinds and 8 right-sizing rows:

| service | kind |
| --- | --- |
| komodor-agent-admission-controller | Deployment |
| cost-profile-control-plane | Node |
| komodor-agent / -daemon / -metrics | Deployment / DaemonSet / Deployment |
| coredns, kindnet, kube-proxy, local-path-provisioner | Deployment / DaemonSet |

Zero rows with an empty `komodorServiceName`.

An earlier revision of this PR claimed the opposite, on the grounds that `metrics-enricher` resolves pod ownership through resources-api, which is fed by the watcher's pod informer. That path is real but it is the **fallback**, not the primary one: telegraf resolves owner refs locally and tags each metric with `komodor_top_owner_ref`, which metrics-collector decodes (`podOwnerRef.go:17-30`) and writes straight onto the cost row. The enricher only handles pods that arrive un-enriched — `read_unenriched_pods.sql` is exactly that set. So removing the informer removes a fallback, not the attribution path.

## What is *not* affected

Telegraf keeps everything it needs. `clusterrole-daemon-metrics` and `clusterrole-metrics-deployment` both grant core `pods`, `namespaces`, `nodes`, `nodes/stats` and `nodes/proxy` on the same ServiceAccount, gated only on `createRbac` + `capabilities.metrics` — never on `allowedResources`. `COST_RBAC_FLOOR` in the test suite pins exactly that, scoped to the ServiceAccount's *effective* permissions rather than to the k8s-watcher role.

The admission controller, HPA/KEDA right-sizing and per-node pricing are likewise untouched.

## Verification

- `helm template ... --set profile=cost` renders, and **26 objects were accepted by a real API server** (`kubectl apply --dry-run=server` against kind). No empty `resources:` list anywhere — the failure mode komodorio/helm-charts#644 exists to prevent.
- **Default install byte-identical.** The only line that moves between any two renders is `checksum/certs-webhook`, which is a hash of the freshly generated self-signed cert and differs between two renders of *identical* code — demonstrated rather than assumed.
- `values_profile_test.py` + `validations_test.py`: **192 pass**.
- **Verified live**, not only rendered: on a `profile=cost` kind cluster the agent creates zero resource informers, cost attribution returns the 9 named services above, and the watcher reports `resourceInfo.enabled=false` for the first time.

## One test list changed meaning

`WATCHER_READS_KEPT` drops to `["replicasets"]`, and `namespaces` moves into `WATCHER_READS_DROPPED`. Worth a reviewer's eye: `replicasets` is left there because it is the ungated entry that stops the `apps` rule rendering an empty resources list — a structural guarantee, not something the cost flows read. The cost pipeline's actual floor is `COST_RBAC_FLOOR`.

---

# The agent's own pod: `role-agent-self.yaml`

The supervisor watches the agent's own pod and, when the watcher restarts, reads that pod's previous logs and events to send a crash report. All of it is scoped to its own namespace and its own pod name (`crash_report_monitor.go`).

It was relying on grants that belong to unrelated features: `pods/log` comes from the cluster-wide rule gated on `capabilities.logs`, which is about reading *customer* pod logs, and `watch` on pods comes only from the metrics Deployment's ClusterRole. Under `profile=cost` the first is off, so the log fetch returned **403** — and `getPreviousLogsForPod` discarded that error and handed a nil reader to `bufio.NewScanner`, panicking the supervisor. Events were refused the same way, though that path degrades rather than crashing.

A namespaced Role removes the coupling: the agent grants itself `pods` get/list/watch, `pods/log` get, and `events` get/list in the release namespace. On a default install this is a strict subset of what the ClusterRoles already give, so no effective permission changes there.

Verified as the ServiceAccount against a live cluster: own-namespace pod logs succeed, `kube-system` pod logs and `kube-system` secrets are still refused.

**This needs komodorio/komodor-agent#616.** That PR fixes the panic itself, plus a second one where a nil pod-status aggregator crash-loops the agent when `allowedResources.pod` is false — which stage 3 sets. **#648 must not merge before an agent release carries it**, or the profile crash-loops on install.

# `capabilities.resourceInfo.enabled` now takes effect

The key only ever reached the agent's ConfigMap, and the agent merges Komodor's remote config over that at startup — same viper layer, merged last. Remote config is not optional (a failed fetch exits the process, `watcher.go:118-122`) and its global fragment always sets `resourceInfo.enabled: true`, so the chart value was overridden on **every** agent, always. Setting it false, including via the profile, did nothing.

An env var sits in a higher viper layer, so the watcher now gets `KOMOKW_RESOURCE_INFO_ENABLED=false` on an explicit false. Gated on the key rather than on the profile, which makes the key real for anyone who sets it, not only for cost installs.

**The default moves `false` → `true` in the same change, and the two halves are load-bearing together.** At the old default every install would have emitted the env var and taken resource-info away from customers who have it enabled remotely — which is what komodorio/planning#241 rejected. `true` is also the honest default: it is what remote config does for every agent today, so the chart now states what actually runs, and the profile's `false` becomes a real deviation instead of a decorative one.

This is the one place the PR changes a chart default, so it deserves a reviewer's attention. Nothing on the backend reads `capabilities.resourceInfo` from `installed-values` — the readers there are `rbac`, `rbacClusterSyncParams`, `actions` and `admissionController`.

Verified live: the agent's effective config reports `resourceInfo.enabled=false` for the first time, and all three consumers of the flag are closed — the event worker (`resource_info_controller.go:137`), the 10s reconciliation task runner (`resource_info_reconciliation_controller.go:68`), and the websocket-driven `Session.Push` that services backend reconcile requests including for CRD-backed kinds (`reconciliation_result/session.go:71`).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,7 +46,11 @@ steps:
       - |
         set -euo pipefail
         BASE_BRANCH="$${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-master}"
-        git fetch -q origin "$$BASE_BRANCH"
+        # Buildkite clones --depth 1, which implies --single-branch, so the only configured
+        # refspec is the default branch. `git fetch origin <branch>` then writes FETCH_HEAD
+        # but no origin/<branch> ref, and the diff below fails on any PR whose base is not
+        # the default branch (i.e. every stacked PR). Fetch into the tracking ref explicitly.
+        git fetch -q origin "+refs/heads/$$BASE_BRANCH:refs/remotes/origin/$$BASE_BRANCH"
         git fetch -q --unshallow 2>/dev/null || git fetch -q --deepen=100 || true
         # Capture diff outside the `if` so `set -e` reaches a `git diff` failure
         # (inside an `if` condition, set -e is disabled and the failure would be masked).
--- a/.buildkite/tests/validations_test.py
+++ b/.buildkite/tests/validations_test.py
@@ -114,7 +114,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_missing_admission_controller_sa_name_when_not_creating(self):
         """Test that admission controller serviceAccount.name is required when create is false"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=true " \
+                   f"--set capabilities.admissionController.enabled=true " \
                    f"--set components.admissionController.serviceAccount.create=false"
         output, exit_code = helm_agent_template(settings=settings)
 
@@ -126,7 +126,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_admission_controller_sa_name_provided_when_not_creating(self):
         """Test that providing admission controller serviceAccount.name when create is false works"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=true " \
+                   f"--set capabilities.admissionController.enabled=true " \
                    f"--set components.admissionController.serviceAccount.create=false " \
                    f"--set components.admissionController.serviceAccount.name=my-ac-sa"
         output, exit_code = helm_agent_template(settings=settings)
@@ -137,7 +137,7 @@ class TestAdmissionControllerServiceAccountValidation:
     def test_admission_controller_disabled_skips_validation(self):
         """Test that admission controller SA validation is skipped when disabled"""
         settings = f"--set apiKey={API_KEY} --set clusterName={CLUSTER_NAME} --set site=us " \
-                   f"--set components.admissionController.enabled=false " \
+                   f"--set capabilities.admissionController.enabled=false " \
                    f"--set components.admissionController.serviceAccount.create=false"
         output, exit_code = helm_agent_template(settings=settings)
 
--- a/.buildkite/tests/validations_test.py
+++ b/.buildkite/tests/validations_test.py
@@ -230,6 +230,15 @@ class TestAllowedResourcesEmptyRuleValidation:
                 "serviceAccount": "serviceaccounts",
             },
         ),
+        "apps": (
+            ["apps"],
+            {
+                "controllerRevision": "controllerrevisions",
+                "daemonSet": "daemonsets",
+                "deployment": "deployments",
+                "statefulSet": "statefulsets",
+            },
+        ),
         "batch": (
             ["batch"],
             {"cronjob": "cronjobs", "job": "jobs"},
--- a/.buildkite/tests/values_components_test.py
+++ b/.buildkite/tests/values_components_test.py
@@ -1,8 +1,9 @@
 import pytest
+import yaml
 from pathlib import Path
 
 from config import RELEASE_NAME
-from helpers.helm_helper import get_yaml_from_helm_template
+from helpers.helm_helper import get_yaml_from_helm_template, helm_agent_template
 from helpers.utils import get_filename_as_cluster_name
 
 CLUSTER_NAME = get_filename_as_cluster_name(__file__)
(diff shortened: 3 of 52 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbexkdp](https://app.clickup.com/t/86cbexkdp)

MorLiberty (approved):

PR #650, opened 2026-09-08
CU-86cbfe87a - fix(komodor-agent): keep the Windows telegraf daemonset under profile=cost

Moran-k:
The cost profile turned off the Windows daemonset. That daemonset is telegraf: the same binary and
the same telegraf.conf as the Linux one, just for Windows nodes.

```
metrics                C:/telegraf/telegraf.exe --config C:/telegraf/telegraf.conf
telegraf-init-sidecar
init container
```

There is no non-metrics container in it. So a `profile=cost` install on a mixed cluster collected
nothing from its Windows nodes, which is the one thing the profile exists to do.

It came in with the original profile PR (#645), in the same hunk as the tunnel and kubectl-proxy
lines, and was the only line in that block with no rationale next to it. It got grouped with
components a cost install does not need without anyone checking what it runs.

## The change

Stop pinning `components.komodorDaemonWindows.enabled`. Its chart default is already `true`, and an
operator on a Linux-only cluster can still set it false, which is what the key is documented for.
This follows rule 2 in the helper's own header: a profile only turns capabilities off, and anything
it depends on staying on is left at the chart default.

## Tests

`test_cost_profile_keeps_every_telegraf_daemonset` asserts both daemonsets render under
`profile=cost`. `test_cost_profile_windows_daemonset_still_runs_telegraf` asserts the metrics
container is still telegraf, since a rendered DaemonSet proves nothing if the container dropped out.
Both fail if the pinning line is put back.

`test_cost_profile_drops_the_workloads_it_disables` had the Windows daemonset in its `gone` list.
Removed, it was asserting the bug.

`values_profile_test.py` 135 passed. Ran the template and values suites locally: 307 passed. The two
failures there are `basic_test.py::test_get_deploy_event_from_resources_api`,
`values_capabilities_test.py::test_get_metrics` and
`values_base_test.py::test_use_existing_service_account`, all live-cluster tests that install into
kind and query the backend, so they need CI.

## Also in here

One stale comment in the same file. #648 added `pod` and `namespace` to the off-list but only
removed `node` from the sentence above it, so the comment claimed two kinds were kept on that the
code four lines below turns off. `values_profile_test.py` already had it right
(`ALLOWED_RESOURCES_ON = ["metrics"]`).

README says the profile turns off the Windows daemonset. Updated, in both `README.md` and
`README.md.gotmpl`.

### DIFF
--- a/.buildkite/tests/values_profile_test.py
+++ b/.buildkite/tests/values_profile_test.py
@@ -33,7 +33,6 @@ PROFILED_CAPABILITY_PATHS = [
     "capabilities.tunnel.kubeapiserver.enabled",
     "capabilities.kubectlProxy.enabled",
     "capabilities.klaudiaIntegrationSync.enabled",
-    "components.komodorDaemonWindows.enabled",
 ]
 
 ALLOWED_RESOURCES_OFF = [
@@ -401,7 +400,6 @@ class TestCostProfile:
 
     def test_cost_profile_drops_the_workloads_it_disables(self, cost_render, default_render):
         gone = [
-            ("DaemonSet", f"{FULLNAME}-daemon-windows"),
             ("ClusterRole", f"{FULLNAME}-node-enricher"),
             ("Service", f"{FULLNAME}-otel-collector"),
         ]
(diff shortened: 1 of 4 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbfe87a](https://app.clickup.com/t/86cbfe87a)

MorLiberty (approved):

PR #651, opened 2026-09-09
CU-86cbfp653 - fix(komodor-agent): grant binpacking the reads its informers need under profile=cost

Moran-k:
`komodor_binpacking_state` is the one telegraf input on the metrics Deployment that is genuinely
cost-related, and under `profile=cost` it is broken. It runs informers over the cluster's
schedulable state, and every one of them is refused.

Measured on a cost-profile cluster, one hour: 1,243 telegraf error lines, all permission denials.
About 870 of those are this input retrying informers that will never sync. Because they never sync,
the input produces no state at all. This is not log noise, it is a cost input returning nothing.

Effective verbs the agent ServiceAccount held on that cluster, from `kubectl auth can-i`:

```
nodes                                 get list        watch missing
poddisruptionbudgets                  -               all missing
persistentvolumes                     -               all missing
persistentvolumeclaims                -               all missing
jobs, cronjobs                        get             list, watch missing
deployments, daemonsets, statefulsets get             list, watch missing
pods, replicasets                     get list watch  fine
```

## The change

Adds the missing `list`/`watch` to `clusterrole-metrics-deployment.yaml`, which is the role bound to
the ServiceAccount the telegraf metrics pod runs as.

Nothing new is being sanctioned on these clusters. `clusterrole-admission-controller.yaml` already
grants this exact set, and renders in full under `profile=cost`. It is bound to the
`komodor-agent-admission-controller` ServiceAccount rather than `komodor-agent`, so the same reads
are already happening on the cluster, just by the other identity. This copies that rule block minus
the writes: no `patch`, no `create`, no `delete`.

Gating the input off instead would contradict the profile, which deliberately keeps
`capabilities.admissionController` and the binpacking and right-sizing webhooks on.

## Effective permission diff

Computed by rendering both profiles, collecting every ClusterRole bound to the agent
ServiceAccount, and unioning the verbs per (apiGroup, resource):

```
default:  no change at all

cost:     apps/daemonsets            [get]            -> [get, list, watch]
          apps/deployments           [get]            -> [get, list, watch]
          apps/statefulsets          [get]            -> [get, list, watch]
          batch/cronjobs             [get]            -> [get, list, watch]
          batch/jobs                 [get]            -> [get, list, watch]
          core/nodes                 [get, list]      -> [get, list, watch]
          core/persistentvolumeclaims                 -> [get, list, watch]
          core/persistentvolumes                      -> [get, list, watch]
          policy/poddisruptionbudgets                 -> [get, list, watch]
```

A default install is untouched because `clusterrole-k8s-watcher.yaml` already grants all of this
there through `allowedResources`. Only the cost profile changes, and it changes by exactly the nine
kinds the informers open.

## Tests

`COST_RBAC_FLOOR` in `values_profile_test.py` already asserts a floor on the ServiceAccount's
effective permissions rather than on any single role, which is the right shape for this. Extended it
with the nine kinds. Reverting the template fails 9 of those assertions.

`replicasets` is in the list and does not fail on revert, because the k8s-watcher role grants it
ungated. Left in as a floor anyway, since the informer needs it and nothing else states that.

145 passed.

## Two lines outside the strict scope

`core/pods` and `autoscaling/horizontalpodautoscalers` floors. Neither needs a chart change, both are
already granted, and both are invisible dependencies of a cost input: HPA survives the profile only
through `capabilities.cost.hpa` at `clusterrole-k8s-watcher.yaml:455`, and nothing pinned it. Say the
word and I will drop them to keep the diff to binpacking alone.

### DIFF
--- a/.buildkite/tests/values_profile_test.py
+++ b/.buildkite/tests/values_profile_test.py
@@ -65,13 +65,25 @@ ALLOWED_RESOURCES_ON = ["metrics"]
 # where cluster region and the node labels the backend derives the cloud provider from come from.
 # Lose that last one and a customer's enterprise discount silently becomes zero.
 COST_RBAC_FLOOR = [
-    ("", "nodes", {"get", "list"}),
-    ("", "pods", {"get", "list"}),
+    ("", "nodes", {"get", "list", "watch"}),
+    ("", "pods", {"get", "list", "watch"}),
     ("", "namespaces", {"get", "list"}),
     ("", "nodes/stats", {"get", "list"}),
     ("", "nodes/proxy", {"get", "list"}),
     ("metrics.k8s.io", "pods", {"get", "list"}),
     ("metrics.k8s.io", "nodes", {"get", "list"}),
+    ("autoscaling", "horizontalpodautoscalers", {"get", "list"}),
+    # komodor_binpacking_state informers. list+watch or the informer never syncs and the input
+    # produces no state at all, which is what happened for the first three cost releases.
+    ("", "persistentvolumes", {"get", "list", "watch"}),
+    ("", "persistentvolumeclaims", {"get", "list", "watch"}),
+    ("policy", "poddisruptionbudgets", {"get", "list", "watch"}),
+    ("apps", "deployments", {"list", "watch"}),
+    ("apps", "replicasets", {"list", "watch"}),
+    ("apps", "statefulsets", {"list", "watch"}),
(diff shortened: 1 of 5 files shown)


Moran-k:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-86cbfp653](https://app.clickup.com/t/86cbfp653)

MorLiberty (approved):

PR #658, opened 2026-09-22
CU-12493t4jnpy - fix(komodor-agent): apply customCa to the admission controller

orsher:
## Problem

`charts/komodor-agent/templates/admission-controller/deployment.yaml` is the **only** agent component that never renders the custom-CA plumbing. Every other component — watcher `deployment.yaml`, `daemonset.yaml`, `daemonset_windows.yaml`, `deployment_metrics.yaml`, `opentelemetry` — includes `ca-init.container` plus the `custom-ca` / `trusted-ca` volumes.

So on a cluster with `customCa.enabled=true`, the setting silently does nothing for the admission controller. It falls back to the image's stock CA bundle, and behind a TLS-intercepting corporate proxy every outbound call to `app.komodor.com` fails:

```
x509: certificate signed by unknown authority
```

Found on a customer cluster running `customCa.enabled: true` with a corporate MITM CA. Over 12h on a single cluster:

| Endpoint | Failures |
|---|---|
| `POST /api/v1/agents/cluster/authenticate` | 2,830 |
| `POST /api/v1/agents/remote-config/admission` | 1,426 |
| `POST /metrics/api/v1/cost/right-sizing/recommendations` | 714 |

Remote config, cluster auth and right-sizing recommendations were all dead — zero right-sizing mutations on that cluster — and had been for at least the full log-retention window. **The pod stayed `Ready` the whole time** with `restartCount: 0`, because its liveness/readiness probes only hit the local webhook server on `:8443` and never touch the backend. Completely silent failure.

Only components that talk to the Komodor backend are affected, which is why the k8s-watcher on the same cluster connected fine and nothing else logged an x509 error.

## Fix

Add to the admission-controller Deployment, each guarded with `with` so they render nothing when `customCa.enabled=false`:

- `initContainers:` → `ca-init.container` (merges the custom CA into the system bundle in the shared `trusted-ca` emptyDir)
- container `volumeMounts` → `custom-ca.trusted-volumeMounts` (`/etc/ssl/certs/`)
- pod `volumes` → `custom-ca.volume` + `custom-ca.trusted-volume`

13 added lines, no deletions. Mirrors exactly what the five other templates already do.

## Verification

Rendered the whole chart on `origin/master` and on this branch:

- **`customCa.enabled=false`** → output is **identical**, no diff at all.
- **`customCa.enabled=true`** → the only difference is the admission-controller Deployment gaining `init-cert`, the `/etc/ssl/certs` overlay mount, and the two volumes. Purely additive; nothing else in the chart moves.
- `helm lint`: 1 chart(s) linted, 0 chart(s) failed.

## Not in this PR

**`ca-init.container` emits `securityContext:` twice** in the same container (it includes both `komodorAgent.customCaInit.securityContext` and `komodorAgent.container.securityContext`). Pre-existing on master in three workloads; this PR makes it four. It is inert — YAML last-key-wins, and both helpers compute identical fields in every scenario that renders — so it is deliberately left out here and cleaned up in a separate PR to keep this one a minimal outage fix.

**`komodorAgent.proxy-conf`** is also missing from this Deployment, but it is **not** a drop-in: with the default `proxy.komodorOnly: true` it emits `KOMOKW_*` env vars, which are the k8s-watcher's prefix. The admission controller uses stock `net/http`, so it only honours `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`. Adding it blindly would render config that looks applied but is ignored. Tracking separately.

Also worth a follow-up: the admission controller should surface backend connectivity in its readiness or an alert, so this class of failure cannot sit unnoticed for weeks.


orsher:
Opened a new task on ClickUp!
Go check it out and add some info: [CU-12493t4jnpy](https://app.clickup.com/t/12493t4jnpy)

MorLiberty (approved):

